PR #4, opened 2016-05-15
Let pass the csrf token in options

hosmelq:


jbrooksuk:
You probably should have made this in two commits.

src/echo.js
```diff
@@ -24,9 +24,9 @@ class PusherConnector {
     /**
      * Create a fresh Pusher connection.
      */
-    static connect(pusherKey, customOptions = {})
+    static connect(pusherKey, { csrfToken, ...customOptions } = {})
```
taylorotwell: I'm not familiar with this syntax. What does it do?

jbrooksuk: The Spread. http://odetocode.com/blogs/scott/archive/2014/09/02/features-of-es6-part-5-the-spread.aspx

PR #11, opened 2016-06-02
Allow a dot as the namespace root for an event

sebastiandedeyne:
This would allow the use of a dot to indicate that the default namespace shouldn't be prepended, keeping event names more consistent.

`echo.channel('foo').listen('\\Bar')` would be the same as  `echo.channel('foo').listen('.Bar')`.


taylorotwell:
Nice thanks.

PR #12, opened 2016-06-03
Allow different connectors to be used with Echo

tlaverdure:
This pull request allows Echo to be used with different web socket providers. PusherJS as previously supported and Socket.io.

## Summary of Changes:
- Added [Typescript](https://www.typescriptlang.org) as a dev dependency. Typescript is a superset of JavaScript that allows for things like abstract classes, interfaces, type hints, and some other language features that someone coming from Laravel may be used to.
- Added [Rollup JS](http://rollupjs.org/) as a dev dependency. Rollup makes use of babel plugins and has a config file so I removed the babelcli npm dependecy and .babelrc file. This is needed to handle compiling typscript on npm install.
- Seperated classes into ther own files for organization.
- Created an abstract Connector
- Added a Socket.io Connector
- Made updates to the Pusher Connector to extend the Connector
- Presence Channel - removed the `then()` method and went directly to the `here()` method. It appears that the intent was to listen to added and removed events at the same time as listening to updated member events, not sure if this was to solve a corner case but one could do this to accomplish that:
```js
echo.join('presence-channel')
        .here(callback)
        .joinging(callback)
        .leaving(callback);
```
- Updated Echo to be instantiated with an object of options for more flexibility.
- Added a bind() method to the Channel class to reflect the on() method for socket.io

## Dev Tooling

To view Typescript type hint errors you can either use vscode, install one of the plugins for atom or sublime, or use the Typescript cli tool by typing `tsc` and `tsc -w` to watch.

To compile the code run `npm run compile`.

@taylorotwell the only real question that I had is when would the `namespace()` method be used? Let me know if you have any questions for me. Thanks


stygiansabyss:
Any chance of fixing the conflicts?  This is something I desperately need in echo.

taylorotwell:
The `namespace` method would be used by the developer somewhere early in their JS application... like `echo.namespace('AppName')` if they are using the Laravel `app:name` command to set the namespace of their application to something other than `App`.

tlaverdure:
@taylorotwell gotcha. That makes sense now, forgot about the `app:name` command. I'll make an update to add that as a configurable option.

So a developer would do:

```js
var echo = new Echo({
    connector: 'pusher',
    namespace: 'AppName'
});
````

**Comment by riyuk** [comment]
The Headline says it's all about a Connector Interface or allowing multiple connectors in echo. But isn't the Pull Request also changing the whole Dev-Language of Echo to Typescript?
Why haven't you used ES6 instead?

**Comment by tlaverdure** [comment]
Used TypeScript for object oriented features. It's all JavaScript in the end.

**Comment by taylorotwell** [comment]
How hard is it to stick with ES6? Is it a significant difference in the code?

**Comment by tlaverdure** [comment]
Not too hard, but would take some time. The code is ES6 with optional types added. Not using Typescript and its compiler would require the abstract Connector class to be refactored or replaced as you can't use those in a plain ES6 setting. One could use the code from this PR to get started on that.

**Comment by taylorotwell** [comment]
I’m fine with keeping TypeScript personally as long as it’s easy for us to maintain.

**Comment by riyuk** [comment]
I've created a PR for a ES6 port. But I can't test it yet so feel free to do so.
Or merge this if you prefer TS as new default.

**Comment by tlaverdure** [comment]
It should be pretty easy to maintain. 

@riyuk your code doesn't appear to work, you should test it out.

**Comment by taylorotwell** [comment]
@tlaverdure what are the conflicts on this?

**Comment by tlaverdure** [comment]
At a glance there appears to be a couple issues at runtime. The node module resolution is off.

```js
var Echo = require('./echo.js');

console.log(Echo); //resolves to - default: function Echo(options)
``` 
That's should be a simple fix, probably need to revert back to the current babel setup. There also appears to be another issue visible in the console log when I tested it out. `Uncaught TypeError: Echo is not a constructor`. One of the reasons I like TypeScript :)

Wasn't able to test further than that...

taylorotwell:
A couple issues at runtime with your TypeScript implementation? So this isn't ready to merge yet?

tlaverdure:
Oh! I thought you were talking about the other PR. This one is good to go. I guess you were referring to git conflicts. I'll check that out, I see some new commits have been added in the last week or so.

tlaverdure:
@taylorotwell conflicts have been resolved.

SelimSalihovic:
Hey guys, is there an approximation of when we can expect this to get merged? @taylorotwell @tlaverdure 🎆

stygiansabyss:
Not sure if I am doing something wrong, but I am not able to get triggered events from echo.  It seems to be hitting all the methods, but when an event is fired, nothing is happening on the front end.

```javascript
  window.echo = new Echo({
    connector: 'socket.io',
    csrfToken: window.csrfToken,
    host:      window.appUrl + ':3000',
  });

      echo.listen('polls', 'PollWasUpdated', event =>
      {
        console.log('triggered')
        this.polls.$set(event.poll.id, event.poll)
      })
```

I never see triggered.

taylorotwell:
After merging this I'm not longer able to compile Echo at all in my Laravel project that uses Browserify. Should I revert?

taylorotwell:
Browserify via Elixir.

taylorotwell:
This is kinda making me nervous about my ability to maintain something written in TypeScript going forward :/

stygiansabyss:
There was another PR that modified this one to use ES6.  Is it possible that could have some pointers on ways to solve this?

taylorotwell:
Dunno. That PR was also broken.

tlaverdure:
Pulled the package from npm and tried to use it in my test project:

```
[16:10:13] gulp-notify: [Laravel Elixir] Browserify Failed!: 'import' and 'export' may appear only with 'sourceType: module'
```

This should fix that: https://github.com/laravel/echo/pull/14

PR #14, opened 2016-07-02
Removed default export since using modules.export

tlaverdure:
Fixes an issue with loading as a module with browserify.


taylorotwell:
So in Elixir 6.0 we're switching to Webpack. I assume this will work with both?

tlaverdure:
It should as any node module. I can test it out too when elixir 6.0 is ready.

morloderex:
Can this me merged soon, i have pulled it down locally and it compiles with both webpack and browserify just perfect. 

If we don' t have this commit it wont compile.

PR #21, opened 2016-08-12
Check Vue exists correctly

mattmcdonald-uk:
Using `if(Vue)` will complain that Vue is not defined when not available.


morloderex:
@mattmcdonald-uk Shoudn't we check if Vue is an object instead of a method?

mattmcdonald-uk:
@morloderex Typeof Vue was returning function for me, object check returned false.

PR #23, opened 2016-08-13
Fixed some type errors and cleanup

tlaverdure:
Just some cleanup and removed a duplicate method.


taylorotwell:
Why do you access window like an array? Is that recommended instead of using dots?

tlaverdure:
In JavaScript you can access a property of an object by it's name using square brackets. In this case the typescript compiler throws an error because it has no definition of said property at design time even though you may know that does exist at run time. I almost always use dot notation unless the property doesn't already exist on the object.

PR #33, opened 2016-08-29
repair socket io leave method

pderiy:
Hi, Taylor

3 days ago I sent to you PR with unsubscribe socket.io method override.
Method "this.socket.removeAllListeners();" doesn't work like I expected. This method deletes all callbacks from all channels. For example I have a reconnector and channel that listening new orders. When I'll leave from orders channel, this method will also remove a reconnector callbacks.

I wrote a method, that saves a collection of channel callbacks. And when user will try to unsubscrube from a channel, then all callbacks will be also deleted.

Now channel collection looks like: 

`this.channels = {`
`    'channel-name' => [`
`        'channel' => channel_object,`
`        'events' => {`
`             "event-name" => [  functions.... ]`
`        }`
`    ]`
`}`

My PR for now working perfect with "underscore". Of course you can do refactoring and remove underscore. These methods tested and work as expected. Right now, method "leave" works good with 'socket.io'. When user is trying to leave from a channel it unsubscribes and removes all callbacks only from one channel, by using method from "socket.io" library - removeListener( event, callback).

P.S. Sorry for my bad english

### DIFF
--- a/src/channel/socketio-presence-channel.ts
+++ b/src/channel/socketio-presence-channel.ts
@@ -13,7 +13,7 @@ export class SocketIoPresenceChannel extends SocketIoChannel implements Presence
      */
     here(callback): SocketIoPresenceChannel {
         this.on('presence:subscribed', (members) => {
-            callback(members.map(m => m.user_info), this.subscription);
+            callback(members.map(m => m.user_info), this.subscription.socket);
         });
 
         return this;
@@ -27,7 +27,7 @@ export class SocketIoPresenceChannel extends SocketIoChannel implements Presence
      */
     joining(callback): SocketIoPresenceChannel {
         this.on('presence:joining', (member) => {
-            callback(member.user_info, this.subscription);
+            callback(member.user_info, this.subscription.socket);
         });
 
         return this;
@@ -41,7 +41,7 @@ export class SocketIoPresenceChannel extends SocketIoChannel implements Presence
      */
     leaving(callback): SocketIoPresenceChannel {
         this.on('presence:leaving', (member) => {
-            callback(member.user_info, this.subscription);
+            callback(member.user_info, this.subscription.socket);
         });
 
         return this;
--- a/src/channel/socketio-channel.ts
+++ b/src/channel/socketio-channel.ts
@@ -53,10 +53,16 @@ export class SocketIoChannel extends Channel {
      * @param  {Function} callback
      */
     on(event: string, callback: Function) {
-        this.subscription.on(event, (channel, data) => {
+        this.subscription.channel.on(event, (channel, data) => {
             if (this.name == channel) {
                 callback(data);
             }
         });
+
+        if(!this.subscription.events) this.subscription.events = {};
+
+        if(!_.has(this.subscription.events, event)) this.subscription.events[event] = [];
+
+        this.subscription.events[event].push(callback);
     }
 }
--- a/src/channel/socketio-channel.ts
+++ b/src/channel/socketio-channel.ts
@@ -1,4 +1,4 @@
-import { EventFormatter } from './../util';
+import { EventFormatter, ArrayHelper } from './../util';
 import { Channel } from './channel';
 
 /**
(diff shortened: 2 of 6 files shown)


taylorotwell:
Can you describe what you're trying to accomplish here in simple terms? Are you trying to unsubscribe from a single channel? From a single event on that channel?

pderiy:
Answer: I'm trying to unsubscribe from one channel and I'm trying to remove all related callbacks for this channel.

Additional information: I'm usind angular routes. When i'm surfing between view states I'm having another channels. And system has to manage callbacks correctly. 

Without this patch: On unsubscribe from a channel system will remove all callbacks (callbacks from another channels too)).

This patch gives you a correct way to manage your channel subscriptions and callbacks.

I'm not sure: This patch use 'unsercore' and it's not good to use underscore for only 3 calls, and you can rewrite it or add underscore (or lodash) to echo dependencies.

If this answer is not good, I can try to describe it better.

taylorotwell:
Can you rewrite it without using underscore?

pderiy:
done

pderiy:
done

pderiy:
Done.

pderiy:
Done.

pderiy:
@taylorotwell Can you accept this PR? If something wrong, just tell me, and I will fix it. Right now it works without underscore.

src/channel/socketio-presence-channel.ts
```diff
@@ -41,7 +41,7 @@ export class SocketIoPresenceChannel extends SocketIoChannel implements Presence
      */
     leaving(callback): SocketIoPresenceChannel {
         this.on('presence:leaving', (member) => {
-            callback(member.user_info, this.channel);
+            callback(member.user_info, this.channel.socket);
```
taylorotwell: I don't understand why you're changing all of these.

pderiy: I'm just trying to send a socket instead of full channel object. Channel object includes list of subscribed events, and user probably dont need this information as was earlier (before this patch channels didnt have event list information). I just send socket info as was before.
image

pderiy:
Hi, @taylorotwell . I know you have another things to do. But I need this update, because when I'm trying to unsubscribe from a channel, it also removes callbacks from another channels. So this update is important for me and for another 'laravel-echo' users. It is tested in real project and it works good. If you have any questions, I'll be good to answer.  Can you accept this PR?

tlaverdure:
I'm familiar with the situation you are dealing with when working with an SPA. I'll take a second look at this @taylorotwell. 

@pderiy in the mean time, just fork the repo and add your patch, and include locally if it is blocking your progress.

taylorotwell:
This has merge conflicts.

pderiy:
@taylorotwell , @tlaverdure fixed merge conflicts.

1. Latest version of laravel-echo(1.0.2) is not working with laravel-echo-server(0.9.65). I believe it was prepeared for laravel-echo-server(1.0.0).

2. I fixed my merge conflicts fot latest update, but for now I have troubles with running laravel-echo-server v1.0.0.

pderiy:
@taylorotwell , @tlaverdure The problem is not fixed in laravel-echo-server v1.0.2.

U can simply test it.
Add script like.
1. Subscribe 'orders' channel.
2. Unsubscribe 'orders' channel.
3. Subscribe 'orders' channel.
4. Emit this any event.

You will receive data 2 times. Because it's still not fixed. But my PR will fix this problem.

tlaverdure:
@pderiy you are right, there is still a case here that needs to be addressed. I'm going to submit a new PR based on what you've submitted that is a bit more concise. There is another issue that I caught as well that comes with killing listeners from different channels with the same event name. 

Will submit soon and I'll also check if we need this for Pusher, which I assume we will since Pusher doesn't automatically maintain the state of callbacks applied to a channel.

pderiy:
@tlaverdure My patch removes only one callback. This patch is not killing listeners from different channels!

Now laravel-echo have a list of channels and channel-listeners. When you need to unsubscribe from a channel it is going foreach loop and removes callbacks from used channel using "socket.io" method ":
`channel.removeListener(event, callback)`

**it doesn't remove listeners from another channels! It is what you need.**
I think this is a good way to resolve this problem.

@tlaverdure you're right, this problem might be in Pusher. But Pusher might have another structure, so you have to write something else for pusher. But for socket.io this is a good way to remember all callbacks for channel and then remove only needed callbacks from needed channels

pderiy:
Thanks @tlaverdure . Your PR is fixing this problem!

PR #36, opened 2016-09-03
Allow pass custom socketIo options

alexmontoanelli:
Hi,

With this options we are allowed to to use same url - without custom port  to connect on socket.io servers.

Example:

We have a laravel-echo-server ( https://github.com/tlaverdure/laravel-echo-server ) running on  port 3001 listening on localhost.

We have a apache virtual host like this:
http://pastebin.com/SMeD6yaZ

This allow us to forward /socket/socket.io direct to laravel-echo-server
with options to upgrade to websocket too.

On our echo-bootstrap we have:

`
window.Echo = new Echo({

    broadcaster: 'socket.io',

    host: 'http://www.myapp.com/',

    socketIoOptions: { path: '/socket/socket.io' }

});
`

What is the main goal:
Our entire app are running on a single addres, and single port.

### DIFF
--- a/src/connector/connector.ts
+++ b/src/connector/connector.ts
@@ -16,7 +16,8 @@ export abstract class Connector {
         csrfToken: null,
         host: null,
         key: null,
-        namespace: null
+        namespace: null,
+	socketIoOptions: {}
     };
 
     /**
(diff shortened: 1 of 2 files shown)


alexmontoanelli:
Forget....

We can pass this options as a host object...

PR #46, opened 2016-09-11
Socket.io Leave Method Improvements and Channel/Connector Updates

tlaverdure:
@pderiy pointed out some issues with event callbacks not being properly cleared with socket.io. I've revised his code to be a bit more concise. The pusher client appears to take care of this so no additional code is needed there. 

I also updated channels to move their subscribe/unsubscribe logic within themselves so that channels have full control of when they subscribe, unsubscribe, and in socket.io's case, handle the tracking the state of event listeners. 

**The connectors are now only responsible for:**

- Creating a connection with the socket provider.
- Creating the different types of channels.
- Facilitating the leaving of channels.

**Channels are responsible for:**

- Subscribing and any logic required to create a channel instance. 
- Unsubscribing along with any logic that goes along with tearing down a channel.

Also some code cleanup.

Fixes https://github.com/laravel/echo/pull/33.

_No breaking changes to the public API._

### DIFF
--- a/src/channel/pusher-channel.ts
+++ b/src/channel/pusher-channel.ts
@@ -5,13 +5,26 @@ import { Channel } from './channel';
  * This class represents a Pusher channel.
  */
 export class PusherChannel extends Channel {
-
     /**
-     * Channel object.
+     * The name of the channel.
      *
      * @type {object}
      */
-    channel: any;
+    name: any;
+
+    /**
+     * The pusher client instance.
+     *
+     * @type {any}
+     */
+    pusher: any;
+
+    /**
+     * Channel options.
+     *
+     * @type {any}
+     */
+    options: any;
 
     /**
      * The event formatter.
@@ -34,12 +47,12 @@ export class PusherChannel extends Channel {
      * @param  {any} pusher
      * @param  {any}  options
      */
-    constructor(
-        public name: any,
-        public pusher: any,
-        public options: any
-    ) {
+    constructor(name: any, pusher: any, options: any) {
         super();
+
+        this.name = name;
+        this.pusher = pusher;
+        this.options = options;
 
         this.eventFormatter = new EventFormatter;
 
--- a/src/channel/pusher-channel.ts
+++ b/src/channel/pusher-channel.ts
@@ -21,21 +21,51 @@ export class PusherChannel extends Channel {
     eventFormatter: EventFormatter;
 
     /**
+     * The subsciption of the channel.
+     *
+     * @type {any}
+     */
+    subscription: any;
+
+    /**
      * Create a new class instance.
      *
-     * @param  {object}  channel
+     * @param  {object} name
+     * @param  {any} pusher
      * @param  {any}  options
      */
-    constructor(channel: any, options: any) {
+    constructor(
+        public name: any,
+        public pusher: any,
+        public options: any) {
         super();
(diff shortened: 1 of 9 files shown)


src/channel/pusher-channel.ts
```diff
@@ -21,21 +21,52 @@ export class PusherChannel extends Channel {
...
+    subscribe(): any {
+        this.subscription = this.pusher.subscribe(this.name);
+    }
+
+    /**
+     * Unsubscribe from a Pusher channel.
+     *
+     * @return {void}
+     */
+    unsubscribe(channel: string): void {
```
taylorotwell: Why does this accept an argument since we're just going to use this.name?

src/channel/pusher-presence-channel.ts
```diff
@@ -5,16 +5,15 @@ import { PresenceChannel } from './presence-channel';
...
      * Register a callback to be called anytime the member list changes.
      *
      * @param  {Function} callback
      * @return {object} this
      */
     here(callback): PusherPresenceChannel {
-        this.bind('pusher:subscription_succeeded', (data) => {
-            callback(Object.keys(data.members).map(k => data.members[k]), this.channel);
+        this.on('pusher:subscription_succeeded', (data) => {
+            callback(Object.keys(data.members).map(k => data.members[k]), this.name);
```
taylorotwell: I don't know if there is any value in passing this.name here? We already know the name I think. The only reason I was passing the Pusher channel is because it has some extra data on it like how many users are subscribed and all of the current users.

src/channel/pusher-presence-channel.ts
```diff
@@ -27,8 +26,8 @@ export class PusherPresenceChannel extends PusherChannel implements PresenceChan
      * @return {PusherPresenceChannel}
      */
     joining(callback): PusherPresenceChannel {
-        this.bind('pusher:member_added', (member) => {
-            callback(member.info, this.channel);
+        this.on('pusher:member_added', (member) => {
+            callback(member.info, this.name);
```
taylorotwell: Same here. Not sure there is any point in passing name here. The consumer already knows the name.

src/channel/pusher-channel.ts
tlaverdure: You are right. Will remove.

src/channel/pusher-presence-channel.ts
tlaverdure: Ok, I will remove this. Each channel now has a "subscription" which is the root object of the socket from which ever provider. (pusher or socket.io)

src/channel/pusher-presence-channel.ts
tlaverdure: diddo

tlaverdure:
@taylorotwell updated.

src/channel/pusher-channel.ts
```diff
@@ -21,21 +21,52 @@ export class PusherChannel extends Channel {
...
      *
-     * @param  {object}  channel
+     * @param  {object} name
+     * @param  {any} pusher
      * @param  {any}  options
      */
-    constructor(channel: any, options: any) {
+    constructor(
+        public name: any,
+        public pusher: any,
```
taylorotwell: Pusher is passed in here but never used?

PR #49, opened 2016-09-20
Fix .subscription to .socket renaming

1618033:


taylorotwell:
@tlaverdure thoughts?

tlaverdure:
Doesn't really seem like a "fix" –  more like a naming preference. Is there a use case that explains why renaming this makes sense?

1618033:
You're completely right.

PR #50, opened 2016-09-21
Update pusher-channel.ts

erikakinga:


deiucanta:
@tlaverdure if you call the listen method, the event name gets processed by the EventFormatter. Pusher publishes some events that you might want to catch. (https://pusher.com/docs/client_api_guide/client_events#subscription_succeeded)

That is possible only by calling `on()` method since `listen()` messes up the event name.

PR #56, opened 2016-10-12
Updated the way X-Socket-ID header is set

alexgarrett:
This fixes the `X-Socket-ID` header not being set correctly and therefore rendering `->toOthers()` (`dontBroadcastToCurrentUser()`) useless.

### DIFF
--- a/src/echo.ts
+++ b/src/echo.ts
@@ -50,7 +50,7 @@ class Echo {
     registerVueRequestInterceptor() {
         Vue.http.interceptors.push((request, next) => {
             if (this.socketId()) {
-                request.headers.set('X-Socket-ID', _this.socketId());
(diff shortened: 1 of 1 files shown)


src/echo.ts
```diff
@@ -50,7 +50,7 @@ class Echo {
     registerVueRequestInterceptor() {
         Vue.http.interceptors.push((request, next) => {
             if (this.socketId()) {
-                request.headers['X-Socket-ID'] = this.socketId();
+                request.headers.set('X-Socket-ID', _this.socketId());
```
taylorotwell: Why _this?

alexgarrett: My bad, sorted.

FWSimon:
This broke my VueJS too @KKSzymanowski and my colleague

PR #58, opened 2016-10-27
Added global settings (currently primarily for the event-formatter);

happyDemon:
- defaultNameSpace: 'App.Events' - makes it trivial to set a default if a developer uses a different app namespace
 - prefixNamespace: bool(true) - if an app uses custom broadcastAs, no namespace prefixes are required (including having to start the event name with a dot).
 - convertNameSpaces: bool(true) - should dots in the event name be converted to \

All the properties can optionally be defined, with defaults that are set to the current way things work.

Example:

```js
window.Echo = new Echo({
    broadcaster: 'pusher',
    key: 'your-pusher-key'
}, {
    defaultNameSpace: 'Project.Events'
});
```

### DIFF
--- a/src/channel/pusher-channel.ts
+++ b/src/channel/pusher-channel.ts
@@ -47,14 +47,14 @@ export class PusherChannel extends Channel {
      * @param  {object} name
      * @param  {any}  options
      */
-    constructor(pusher: any, name: any, options: any) {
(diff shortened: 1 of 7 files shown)


taylorotwell:
I think the other PR solved this.

Issue #55, opened 2016-10-07
Can subscribe, but callback isn't triggered on event.


cedricve:
I've managed to setup and install Laravel-echo correctly however the callback isn't triggered when an event is send from the debug tool. I see the client connecting succesfully. Can someone help? 

![screen shot 2016-10-07 at 10 54 46](https://cloud.githubusercontent.com/assets/1546779/19184302/c22ae0e6-8c7c-11e6-8ead-60329902ea82.png)

![screen shot 2016-10-07 at 10 54 56](https://cloud.githubusercontent.com/assets/1546779/19184305/c3b8eeda-8c7c-11e6-8026-561478ce5a4e.png)

FourStacks:
I'm having the exact same problem.  First I went through the docs and got my events talking to pusher just fine but my client side code doesn't trigger any callback on the listen method.

To be on the safe side I did a bare bones example following along with Matt Stauffers blogs post:

https://mattstauffer.co/blog/introducing-laravel-echo

Same outcome though.  Works just fine for sending the event to pusher but the JS echo aspect just doesn't fire  the callback.

I switched out laravel echo for the regular pusher library and code and the event is picked up in the browser just fine so I think it's probably an echo issue rather than anything to do with pusher per se.

    import Echo from "laravel-echo"

    // Init Echo
    window.Echo = new Echo({
        broadcaster: 'pusher',
        key : 'pusher-key-redacted',
        encrypted: true,
        cluster: 'mt1'
    });

    window.Echo.channel('chat-room.1')
	.listen('ChatMessageWasReceived', (e) => {
	     console.log('here');
	     console.log(e.test);
	});

For reference the 'test' in the listen callback is the public property on my event class

FourStacks:
@cedricve - I'm guessing you're not getting any other error messages in the console either right?  

What sort of app are you building?  Mine's a single page application built using vue js, vue router and vuex.  Not sure that should make any difference but figured it was worth seeing if there are any similarities/differences in our use cases

cedricve:
Well, it's just a proof-of-concept the code is in the bootstrap.js file, as recommended by the documentation. I was really hoping to get this work.. I think I'll move to the official library as wel. Thanks for the help

FourStacks:
Yeah, bit frustrating as Echo has some features that look pretty neat so I'd like to get it working if possible.  If anyone else has any ideas or is having the same issue then I'd be happy to provide more info on my own usage

FourStacks:
[...]

PR #59, opened 2016-10-27
Allows namespace option to be set to false.

tlaverdure:
**Fixed a typing issue**
`void` → `PusherChannel`

**Allows default namespace to be omitted.**
This is preferred when using laravel-echo with another framework that doesn't use the Laravel namespacing conventions. 

``` javascript
new Echo({
   namespace: false
});
```

``` javascript
echo.channel('test-channel').listen('test-event') //formatted as test-event
```

Closes https://github.com/laravel/echo/issues/55

### DIFF
--- a/src/util/event-formatter.ts
+++ b/src/util/event-formatter.ts
@@ -8,7 +8,7 @@ export class EventFormatter {
      *
      * @type {string}
      */
-    defaultNamespace: string = 'App.Events';
+    defaultNamespace: string | boolean = 'App.Events';
 
     /**
      * Format the given event name.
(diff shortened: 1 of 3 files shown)


src/util/event-formatter.ts
```diff
@@ -17,10 +17,12 @@ export class EventFormatter {
      * @return {string}
      */
     format(event: string): string {
-        if (event.charAt(0) != '\\' && event.charAt(0) != '.') {
-            event = this.defaultNamespace + '.' + event;
-        } else {
-            event = event.substr(1);
+        if (this.defaultNamespace !== false) {
```
happyDemon: this is actually nicer than my pull request https://github.com/laravel/echo/pull/58 😬

PR #62, opened 2016-10-31
Removed unnecessary check

dzava:
Goes with #59.

### DIFF
--- a/src/channel/pusher-channel.ts
+++ b/src/channel/pusher-channel.ts
@@ -55,10 +55,7 @@ export class PusherChannel extends Channel {
         this.options = options;
 
         this.eventFormatter = new EventFormatter;
-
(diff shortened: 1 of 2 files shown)


tlaverdure:
`namespace` is always part of the default options so I don't believe this PR would actually fix anything. If you don't want to use a namespace you will need to set the option to false. 

This also doesn't appear to be published to npm yet, which means it hasn't been made available.

dzava:
You are correct, the check should be removed altogether.

dzava:
Actually we should either change the condition to check for an empty string or change the default namespace in connector.ts.

taylorotwell:
Needs actual explanation of what we're trying to do here. Please re-open a PR with a proper description and explanation if we need to fix something.

PR #68, opened 2016-12-08
Remove duplicate tsconfig entry

kylestev:
Visual Studio Code raises a warning when viewing the tsconfig.json file in this repository.


kylestev:
The first instance of this configuration key can be found on line 6: https://github.com/laravel/echo/blob/fbd14e5b22f1247a30cd87c34af844f5bd31e02b/tsconfig.json#L6

PR #75, opened 2017-01-02
[SocketIO] Subscribe to channel on reconnect

barryvdh:
When a connection is lost (either server is killed or client lost connection) and reconnected, a new socket is created and isn't subscribed.

This adds a reconnect listener to subscribe when a new connection is created. This seems to fix issue https://github.com/tlaverdure/laravel-echo-server/issues/90
I don't know of any other SocketIO servers, so not sure if this is possible to handle server-side otherwise.

/cc @tlaverdure

### DIFF
--- a/src/channel/socketio-channel.ts
+++ b/src/channel/socketio-channel.ts
@@ -64,7 +64,9 @@ export class SocketIoChannel extends Channel {
         this.eventFormatter = new EventFormatter(this.options.namespace);
 
         this.subscribe();
-        socket.on('reconnect', this.subscribe);
+        this.socket.on('reconnect', () => {
+            this.subscribe();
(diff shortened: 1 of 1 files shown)


taylorotwell:
@tlaverdure what do you think of this?

barryvdh:
One potential issue could be that the CSRF token is expired after a given time, while the web app would still be authenticated with cookies. So that would perhaps need a way to refresh the CSRF token?

tlaverdure:
Hey Guys, I got a chance to test this out and it looks like it should work. After killing my echo server and restarting it my channels re-authenticated and I was up and running as usual.

barryvdh:
Thanks!

Issue #47, opened 2016-09-17
any plan to support triggering client events?


shwhl:
any plan to support triggering client events?

morloderex:
Create a pr for it and see :)

tlaverdure:
I may have something in the works in the next week or so.

shigg1120:
nice one

luiz-pv9:
This would be really helpful. There are actions that don't need to go through Laravel, such as "user is typing..." in a chat, for example. This is just a simple broadcast to a channel.

woganmay:
Any news on this? I've got Echo working for receiving messages to the client, but sending stuff back from the client would be very helpful.

lukepolo:
if your using pusher you can just use their native event triggers : 

https://pusher.com/docs/client_api_guide/client_events#trigger-events

woganmay:
@lukepolo Thanks for the link! I'm using socket.io for now, but to be honest, looking at Pusher's pricing I can probably afford it for the app I plan on building.

lukepolo:
@woganmay Gotcha, i dont think it would be terribly difficult to implement, but I do not have the time currently. But will need it for my app soon. So hopefully work will pay me to implement this soon!

shigg1120:
The workaround for this is just make another instance of a channel and use "trigger" method right?If you are using Pusher right?

Does Echo have a method like it?

tillkruss:
```js
var channel = Echo.join('foobar');

channel.here((users) => {
    channel.pusher.channels.channels[channel.name].trigger('client-foobar', { foo: 'bar' });
});
```

shwhl:
Closing the issue, the `whisper` method and `listenForWhisper` method is available now.

brunoinds:
Whisper does not solve the problem. It will be good to implement in the Laravel Echo library some way to trigger Events on the server. I know is not possible because Laravel does not listen for WS, but the JS under the hood could send a Post request....

PR #79, opened 2017-01-06
Support client events on Pusher presence channels

tillkruss:
This PR adds a `trigger()` method onto the `PusherPresenceChannel` to trigger [a client event](https://pusher.com/docs/client_api_guide/client_events#trigger-events).

This allows the browser to send events directly to Pusher, without having to go through Laravel. It's useful for events such as a "typing..." indicator or other events where it's overkill to pass them through Laravel. In my case it's when a user starts/stops broadcasting video.

See also: https://github.com/laravel/echo/issues/47

### DIFF
--- a/src/channel/pusher-presence-channel.ts
+++ b/src/channel/pusher-presence-channel.ts
@@ -46,4 +46,16 @@ export class PusherPresenceChannel extends PusherChannel implements PresenceChan
 
         return this;
     }
+    
+    /**
+     * Trigger client event on the channel.
+     *
+     * @param  {Function}  callback
+     * @return {PusherPresenceChannel}
+     */
+    trigger(eventName, data): PusherPresenceChannel {
+        this.pusher.channels.channels[this.name].trigger(eventName, data);
+
+        return this;
+    }
 }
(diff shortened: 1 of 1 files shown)


taylorotwell:
Should client- automatically be prefixed to the event name? Pusher requires them to start with that prefix.

tillkruss:
Sure, done!

barryvdh:
Should we also implement this for laravel-echo-server and add it to the interface? (cc @tlaverdure )

taylorotwell:
I'm curious why this wasn't done on Private channels as well? Why only PresenceChannels?

tillkruss:
Because there is no `pusher-private-channel.ts` and I'm not quite sure how to add that file.

taylorotwell:
OK I'll look into adding that when I merge this.

tillkruss:
Thanks a lot!

taylorotwell:
Added support for private channel.... renamed "trigger" to "whisper"... added "listenForWhisper" to help with listening for them.

Released as Echo 1.2.0.

PR #82, opened 2017-01-10
Optimize csrfToken()

tinpont:
In some situation, we don't have `window` or `document` object, such as `node.js`, `react-native` etc. So we need to detect if the object exists.


taylorotwell:
Why even let select = null? Can't you just remove that whole line at this point?

tinpont:
Sure, done.

PR #86, opened 2017-01-12
Pass full options to socketIO

### LINKED ISSUE #80: Ability to pass on option "path" to socket io connection
Please allow passing on path in options to socket io connector.

It is not possible to use different route than "/" because anything after "/" is treated as namespace.

e.g.
var echo = new Echo({
    broadcaster: 'socket.io',
    host: 'http://localhost',
    path: '/myapp/socket.io/'
});

This path should be used while creating io socket as:

this.socket = io(this.options.host, {path: this.options.path});

Thanks very much in advance.

barryvdh:
Pass entire options to the options parameter to allow customisation, similar to Pusher.
Fixes #80

### DIFF
--- a/src/connector/socketio-connector.ts
+++ b/src/connector/socketio-connector.ts
@@ -25,7 +25,7 @@ export class SocketIoConnector extends Connector {
      * @return void
      */
     connect(): void {
-        this.socket = io(this.options.host);
+        this.socket = io(this.options.host, this.options);
 
         return this.socket;
     }


taylorotwell:
Does still work if someone specifies options socket.io doesn't support?

barryvdh:
I think so. `host` and `broadcaster` are already options that don't exist in SocketIO. Only problem could be if they DO exist, because of conflicts..

taylorotwell:
Did you test it?

barryvdh:
Yes I tested that the options are passed to the io instance, and that it still works.

PR #87, opened 2017-01-12
[SocketIO] Implement whisper for SocketIO

barryvdh:
Implement 'whisper' functionality for SocketIO, based on Pusher.
 
 - Move listener to basic channel
 - Create private SocketIO channel with whisper method
 - Extend Presence channel from private channel

cc @tlaverdure for server implementation, serverside PR: https://github.com/tlaverdure/laravel-echo-server/pull/100

### DIFF
--- a/src/channel/channel.ts
+++ b/src/channel/channel.ts
@@ -28,4 +28,15 @@ export abstract class Channel {
     notification(callback: Function): Channel {
         return this.listen('.Illuminate.Notifications.Events.BroadcastNotificationCreated', callback);
     }
+
+    /**
+     * Listen for a whisper event on the channel instance.
+     *
+     * @param  {string} event
+     * @param  {Function}   callback
+     * @return {PusherChannel}
+     */
+    listenForWhisper(event: string, callback: Function): Channel {
+        return this.listen('.client-' + event, callback);
+    }
 }
--- a/src/channel/index.ts
+++ b/src/channel/index.ts
@@ -4,4 +4,5 @@ export * from './pusher-channel';
 export * from './pusher-private-channel';
 export * from './pusher-presence-channel';
 export * from './socketio-channel';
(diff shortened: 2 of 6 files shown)


barryvdh:
Note that this was just an implementation to see if it works correctly, we need to add verification for the client. Just check if the socketId exists in `this.io.sockets.adapter.rooms`?

barryvdh:
Updated event name to be more clear: `client event`

taylorotwell:
If it has to hit the server is it really a counterpart to the Pusher version?

taylorotwell:
I guess it doesn't have to hit the Laravel application?

barryvdh:
It doesn't hit the Laravel server I think, it emits a message to the Node server and the broadcasts is to the other clients. See https://github.com/tlaverdure/laravel-echo-server/blob/f1ce0fca5c7198af9d9befc98483fd7488d9adfc/src/channels/channel.ts#L69-L81

barryvdh:
Or where do you see it hitting Laravel?

tlaverdure:
Looks good to me 👍

taylorotwell:
Thanks :heart:

zdenektopic:
Is there also something at actually sends events from client to laravel?
Example: Chat application
- users sends message over socket.io
- node hits laravel
- laravel receives event with user and channel info

I guess its the same as making HTTP request directly to laravel, but can you do that simply with context of the user AND CHANNEL?

barryvdh:
What would be the benefit over a plain XHR request to Laravel directly?

zdenektopic:
You are right, there is not actual benefit. It just makes it possible to wrap it all in one communication layer.
Anyway, is it possible to send XHR that resolves the channel and user context in laravel?

Actually the Echo JS client could be able to emit events back to Laravel directly with XHR.

PR #102, opened 2017-02-17
Add sanity check for socketId

inkadnb:
This fixes the jQuery interceptor. We don't want to add the X-Socket-Id header if the socketId is undefined.

This bug managed to break the Google Maps API through jQuery.

### DIFF
--- a/src/echo.ts
+++ b/src/echo.ts
@@ -85,7 +85,9 @@ class Echo {
         if (typeof jQuery.ajax != 'undefined' ) {
             jQuery.ajaxSetup({
                 beforeSend: (xhr) => {
-                    xhr.setRequestHeader('X-Socket-Id', this.socketId());
+					if (this.socketId()) {
+						xhr.setRequestHeader('X-Socket-Id', this.socketId());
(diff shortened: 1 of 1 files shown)


bbashy:
I'm getting this error and just wondering if it's related?
```
XMLHttpRequest cannot load https://maps.googleapis.com/maps/api/geocode/json?address=someaddress&key=keyhere.
Request header field X-Socket-Id is not allowed by Access-Control-Allow-Headers in preflight response.
```

morloderex:
@bbashy It might be because jquery.ajaxSetup is global setup of the .ajax method....

bbashy:
@morloderex And so now every ajax call is now including X-Socket-Id header?

I tried commenting out the part where it adds it to the headers and it was still happening.

morloderex:
@bbashy It seems like it. yes.

bbashy:
Getting this as well just because Echo is loaded.
> Request header field Access-Control-Allow-Origin is not allowed by Access-Control-Allow-Headers in preflight response

PR #106, opened 2017-03-10
Support react-native variant of pusher

mattmcdonald-uk:
Pusher has a react-native variant which works with RN (default doesn't - tries to use document object).

Also use a more strict document check to avoid a thrown error in RN as document isn't available.

### DIFF
--- a/src/connector/connector.ts
+++ b/src/connector/connector.ts
@@ -65,7 +65,7 @@ export abstract class Connector {
             return window['Laravel'].csrfToken;
         } else if (this.options.csrfToken) {
             return this.options.csrfToken;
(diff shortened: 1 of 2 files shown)


ianlandsman:
I'm getting the error described here and this commit seems highly suspicious! https://laracasts.com/discuss/channels/general-discussion/echo-build-failing-due-to-pusherjs-requiring-react-native

ianlandsman:
Further update, forcing Echo 1.2.3 fixes the issue.

tlaverdure:
This should fix for webpack - https://github.com/laravel/echo/pull/107

Issue #40, opened 2016-09-07
Requires pusher-js if using socket.io


petecoop:
I'm getting `Cannot find module 'pusher-js'` when packing with browserify and need to require it even when using socket.io.

Could it be moved to outside in a similar way to socket.io? I'd imagine there would be other connectors in the future, and think it would be best if you just install the library for the connector you are using.

petecoop:
As a quick workaround for this I'm using `.ignore('pusher-js)` in my browserify setup in gulp which just replaces it with an empty object. But would be nice to not have to do this

Issue #52, opened 2016-09-22
[Suggestion] remove pusher and rollup from dependency list


ctf0:
1- for users using socket.io , u still have to install pusher.js or u would get an error https://github.com/laravel/echo/issues/40

``` js
ERROR in ./~/laravel-echo/dist/echo.js
Module not found: Error: Can't resolve 'pusher-js'
```

2- elixir is using webpack which mostly is what users will be using, so no need for rollup

in both cases it should be left to the user to decide which tool he/she will be using.

**UPDATE**
https://github.com/laravel/echo/pull/110

seancheung:
Temporarily  you may use webpack config to ignore *pusher-js*:

Add this file to your project root:

> webpack.config.js

```js
var webpack = require("webpack");

module.exports = {
    plugins: [
        new webpack.IgnorePlugin(/pusher-js/),
    ]
}
```

AliBoukthir:
@seancheung thank you so much .

PR #107, opened 2017-03-13
Dynamic requiring of pusher libraries.

tlaverdure:
By default the [pusher-js ](https://github.com/laravel/echo/blob/99be9119903d679980cadb4d6758cfa9c3b244d5/src/echo.ts#L46) and [pusher/react-native](pusher/react-native) packages are required if the pusher broadcaster is selected. This is causing Webpack to fail if the packages are not installed. I'm opposed to installing additional packages when they are not is use, as that potentially leads to a large bundle size. Webpack as a module bundler does not have access to the `window` object at build time  to determine if these packages should be required or not. 

This is the best solution that I could find and it seems to work as intended.

**Reference:**
https://webpack.github.io/docs/context.html#dynamic-require-rewriting 

Closes https://github.com/laravel/echo/issues/40, https://github.com/laravel/echo/issues/52

### DIFF
--- a/src/echo.ts
+++ b/src/echo.ts
@@ -43,16 +43,20 @@ class Echo {
 
         if (this.options.broadcaster == 'pusher') {
             if (! window['Pusher']) {
-                window['Pusher'] = require('pusher-js');
+                let pusherJs = 'pusher-js';
+
+                window['Pusher'] = require('' + pusherJs);
             }
 
             this.connector = new PusherConnector(this.options);
         } else if (this.options.broadcaster == 'pusher/react-native') {
-                if (! window['Pusher']) {
-                    window['Pusher'] = require('pusher-js/react-native');
-                }
+            if (! window['Pusher']) {
+                let pusherReactNative = 'pusher-js/react-native';
 
-                this.connector = new PusherConnector(this.options);
+                window['Pusher'] = require('' + pusherReactNative);
+            }
+
+            this.connector = new PusherConnector(this.options);
         } else if (this.options.broadcaster == 'socket.io') {
             this.connector = new SocketIoConnector(this.options);
         }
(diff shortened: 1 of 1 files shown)


mKeRix:
This probably fixes #108 as well!

lukepolo:
This did not fix for me still getting 

`Cannot find module 'pusher-js'.`

tlaverdure:
Are you using webpack or a different bundler?

lukepolo:
webpack

aaronhuisinga:
I'm having the same issue as @lukepolo on a standard laravel-mix compilation. I had to revert back two versions as if I only reverted one, I had the cannot find pusher-react error that this fixed.

P9Q:
Uncaught Error: Cannot find module 'pusher-js'.

Laravel version 5.4.15

taylorotwell:
😦 can I just revert these changes as we obviously don't know how to fix this right now?

tlaverdure:
Well it works for my Webpack configuration, so it does resolve some of the issues, but I think that it's not a 100% fix based on the different JS bundling setups devs may be using. This issue has been existent for a while, since pusher-js was added as a dependency, and now with pusher/react-native added, I feel like a precedent has been established that these 3rd party libs should be distributed with laravel-echo.

To fix I would suggest removing pusher-js and pusher/react-native from this project and documenting that those files be required in the project they are being used before instantiating laravel-echo.

taylorotwell:
Can someone make whatever PR we need in order to make Echo not broken?

tlaverdure:
Sure I'll send in a sec.

P9Q:
@tlaverdure My problem solved, as you mentioned here :point_right: [#110](https://github.com/laravel/echo/pull/110), I just added following line:

`window.Pusher = require('pusher-js');`

into "resources\assets\js\bootstrap.js" and it's working fine,
Thank you and @taylorotwell :heart_eyes: :+1: :+1: :ok_hand:

PR #110, opened 2017-03-14
Remove the pusher libraries.

tlaverdure:
Requiring pusher libraries directly in the laravel-echo is not the best strategy in terms of using pusher, socket.io, or another broadcaster. This PR removes the hardcoded dependency of the pusher client libraries and requires that these packages be added during configuration.  

**This is a small breaking change in terms of setting up echo.**

This will also need to be updated here:
https://laravel.com/docs/5.4/broadcasting#configuration

``` js
import Echo from "laravel-echo"

window.Pusher = require('pusher-js');

window.Echo = new Echo({
    broadcaster: 'pusher',
    key: 'your-pusher-key'
});
```

### DIFF
--- a/src/echo.ts
+++ b/src/echo.ts
@@ -42,20 +42,6 @@ class Echo {
         }
 
         if (this.options.broadcaster == 'pusher') {
-            if (! window['Pusher']) {
-                let pusherJs = 'pusher-js';
-
-                window['Pusher'] = require('' + pusherJs);
-            }
-
-            this.connector = new PusherConnector(this.options);
-        } else if (this.options.broadcaster == 'pusher/react-native') {
-            if (! window['Pusher']) {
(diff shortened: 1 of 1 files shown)


lukepolo:
I quite like this PR , thanks for this (also think socket.io please will be really pleased not to have to pull in another library)

aaronhuisinga:
This makes a ton of sense to me. Great PR.

paulgrimaldo:
thanks for help.

andrewmclagan:
Why is it being bound to `window. this is not a good idea. There are many server rendered use cases where `window` is not available. Assuming this will be used in a browser is erroneous.

edwardhalley:
Hi How about ionic laravel echo ?

Bilal-Basharat:
in electron js, it is giving the error 'window is not defined'

Issue #94, opened 2017-01-22
Can't disconnect from Pusher


elasticsteve:
Pusher has the disconnect() method. Is there such a thing in Echo? I couldn't see it. It's very important for me to disconnect users who keep their tab(s) open otherwise I run out of connections.

RusinovAnton:
You can use ".leave()" method which is described in the [doc](https://laravel.com/docs/5.3/broadcasting)

elasticsteve:
But that's not the same. You only leave a channel with this, which still keeps the Pusher connection open, I saw. Only pusher.disconnect() will close it, which is not available through Echo.

RusinovAnton:
@elasticsteve It would be really useful if you could give your use case as example.

elasticsteve:
I said it in my first comment: to disconnect users keeping their tabs open. It's a common problem with Pusher, because it eats up all your available connections and you max out.

RusinovAnton:
@elasticsteve I mean, could you write some pseudo code implementation of that. How would you use  .disconnect() if it were implemented. Will you need .reconnect() ?
The approach I'm thinking on right now is to disconnect from Pusher on .leave() call without channel name argument
```
Echo.leave() // this.pusher.disconnect()
Echo.leave('my-channel') // his.channels[name].unsubscribe()
```
Whats your opinion on this?

elasticsteve:
You open a connection with new Echo();

At that moment Pusher shows an active connection, even if there is no listener subscribed. We need something to close that connection again, so Pusher releases the connection. On my account I have a max of 200 connections, but if people just leave their tabs open on a computer that is on all day or at work, I run out of connection within hours. Pusher doesn't disconnect. Even worse if one user has my app on tabs, they eat up 4 connections! You can see how soon you run out.

I think Echo. disconnect(); would be brilliant for that, similar to pusher.disconnect()

RusinovAnton:
Ok Im going to show you how I use Echo with ReactJS for my app and why I'm concerned about adding .disconnect() wouldn't be that easy as it seems:
```
// AppEchoInstance.js

import Echo from 'laravel-echo'

export default new Echo({/* my app pusher config */})


// NotificationsComponent.js

import AppEchoInstance from 'AppEchoInstance'

...
//
componentDidMount() {
  // Using single instance here I'm having only one connection per page, not multiple if I'm not wrong
  AppEchoInstance
    .channel('my-channel-1')
      .listen('event1', () => {})
      .listen('event2', () => {})
[...]

PR #125, opened 2017-05-28
Added a function to disconnect from server

maxpaynestory:
Added the functionality to disconnect from socket.io or pusher servers. This feature was requested on #94

### DIFF
--- a/src/echo.ts
+++ b/src/echo.ts
@@ -10,9 +10,9 @@ class Echo {
     /**
      * The broadcasting connector.
      *
-     * @type {Connector}
+     * @type {object}
      */
-    connector: Connector;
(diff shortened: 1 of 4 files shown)


src/echo.ts
```diff
@@ -10,9 +10,9 @@ class Echo {
     /**
      * The broadcasting connector.
      *
-     * @type {object}
+     * @type {Connector}
      */
-    connector: any;
+    connector: Connector;
```
taylorotwell: This change seems irrelevant? Please revert this and the above change.

maxpaynestory: I have reverted the irrelevant part.

Issue #119, opened 2017-04-24
Laravel broadcastAs and Echo listen mismatch


yazfield:
Hi,

Laravel gives the possibility to name event's broadcast, however, Echo assumes that the event's name is always a namespace.
Following this [documentation](https://laravel.com/docs/5.4/broadcasting#broadcast-name):
```php
public function broadcastAs()
{
    return 'server.created';
}
```
I was trying to broadcast the event via Pusher, and had this code in my Javascript:
```javascript
window.Echo = new Echo({
    // settings
});
window.Echo.private(`App.User.${user.id}`)
.listen('.server.created', (e) => {
    console.log('notif arrived', e);
})
```
But wasn't receiving any notifications in the front end, the problem was in Echo's `event-formatter`, the format method replaces all `.` with `\\` 
```javascript
format(event: string): string {
        if (this.namespace) {
            if (event.charAt(0) != '\\' && event.charAt(0) != '.') {
                event = this.namespace + '.' + event;
            } else {
                event = event.substr(1);
            }
        }
        // Why is this here?
        return event.replace(/\./g, '\\');
    }
```

Now I'm not sure if the problem is:
1. is it in Laravel Event `broadcastAs` method that should format the event's name similar to Echo?
1. or Echo shouldn't always format the event name and allow the developer to choose? which seems to be the right way to do it, unless I'm missing something. I can PR this.
3. or Laravel documentation is incorrect?
4. or I'm missing something.

Thanks

hernandev:
made me crazy for a couple of hours.

ellisio:
Seeing as how `namespace` is exposed in Echo's options, it might be worth changing the documentation to reflect that doing the following will prevent `App.Event.` from being prepended.

```
let Echo = new Echo({
    namespace: null
});
```

Then changing the formatter to the following will be enough to fix this issue.

```
format(event: string): string {
    if (this.namespace) {
        event = this.namespace + '.' + event;
    }

    return event;
}
```

This will allow a developer to do the following successfully:

```
// PHP
broadcastAs() {
    return 'server.created';
}

// JS
.listen('service.created', () => {});
```

driesvints:
This just works for me so I believe this is fixed in the meantime?

gabrieladeniji:
> broadcastAs() { return 'contact.created'; }

It worked for me, Thanks

PR #133, opened 2017-07-25
Long term fix for dot syntax & namespace syntax.

ellisio:
This is the alternative longterm fix for #119. The temporary fix is #132.

Refer to https://github.com/laravel/echo/issues/119#issuecomment-317627591 for the explanation on how this fix can be used.

Created two PRs to let the maintainers determine the best course of action.

### DIFF
--- a/src/connector/connector.ts
+++ b/src/connector/connector.ts
@@ -16,7 +16,7 @@ export abstract class Connector {
         csrfToken: null,
         host: null,
         key: null,
-        namespace: 'App.Events'
+        namespace: 'App\\Events'
     };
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
Close one of your PRs. I would prefer to have one proper fix for whatever the problem is.

ellisio:
@taylorotwell This is the preferred fix for long term. #132 would introduce a new option to allow dot syntax to work, and be backwards compatible. This new version is _not_ backwards compatible, but a better fix to sync up event naming in Echo to how Laravel does it within it's core.

PR #134, opened 2017-07-25
fixed the dotsyntax bug

ellisio:
Okay, it is a wonder what sleep will do to the brain. All that is needed to fix `server.created` `broadcastAs()` methods was to move the `.replace()` call. I tested this fix with the following configurations:

```
// PHP
namespace App\Events;
class ServerCreated extends Event {}

// JS
let Echo = new Echo();
Echo.listen('ServerCreated', () => {});
```

```
// PHP
namespace App\Events\Server;
class Created extends Event {}

// JS
let Echo = new Echo();
Echo.listen('Server.Created', () => {});
```

```
// PHP
namespace Foobar\Events;
class ServerCreated extends Event {}

// JS
let Echo = new Echo();
Echo.listen('\\Foobar\\Events\\ServerCreated', () => {});
Echo.listen('.Foorbar.Events.ServerCreated', () => {});
```

```
// PHP
namespace App\Events;
class ServerCreated extends Event {
    public function broadcastAs() { return 'server.created'; }
}

// JS
let Echo = new Echo();
Echo.listen('.server.created', () => {});
```

All of these scenarios triggered their event handlers as expected.

### DIFF
--- a/src/util/event-formatter.ts
+++ b/src/util/event-formatter.ts
@@ -26,12 +26,10 @@ export class EventFormatter {
      * @return {string}
      */
     format(event: string): string {
-        if (this.namespace) {
-            if (event.charAt(0) != '\\' && event.charAt(0) != '.') {
-                event = this.namespace + '.' + event;
-            } else {
-                event = event.substr(1);
-            }
+        if (event.charAt(0) === '.' || event.charAt(0) === '\\') {
+            return event.substr(1);
+        } else if (this.namespace) {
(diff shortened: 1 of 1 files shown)


lukepolo:
For some reason this broke notifications , 
`Pusher : No callbacks on private-App.Models.Site.Site.3 for Illuminate\Notifications\Events\BroadcastNotificationCreated`

When i revert to 1.3.0 , things work fine. I dont have the time to fix it now , but figured id let you know

Issue #135, opened 2017-07-26
Not receiving notification 1.3.1


msonowal:
After I updated to v1.3.1 it my echo on client side not working i.e. I was listening to private channel
on notification which is not getting executed.
which was working fine previously. below is the code
`Echo.private('user.' + User.id)`
`.notification((notification) => {`
`   console.debug('Notification received ', notification);`
`});`
but after digging for sometime if I do something like this I can get the event but it feels dirty

 ` .listen('Illuminate\\Notifications\\Events\\BroadcastNotificationCreated', (e) => {`
`     console.log('Event Notification received ', e)`
`})`

and here is my EchoInstance

`const EchoInstance = new Echo({`
`    broadcaster: 'socket.io',`
`    host: window.location.hostname + ':6001',`
`    namespace: 'booking_app.Events',`
`});`

indigoram89:
same problem in 1.3.1:

event comes to client:
![2017-07-27_17-27-31](https://user-images.githubusercontent.com/10687211/28675597-04a10a62-72f1-11e7-87b9-d7ca25137707.png)

but callback not called:
![2017-07-27_17-29-06](https://user-images.githubusercontent.com/10687211/28675633-1e71c238-72f1-11e7-8793-c45e7febcb7a.png)

nguyenhaiphan:
I don't know why but I got this too, maybe the temporarily workaround is to stay with the old version of this package.

SergeyMiracle:
Same here, 1.3.0 not working

lukepolo:
I had to revert to 1.3.0 as well, looking into what changed in 1.3.1 now

sietzekeuning:
I also had to change the listener to:
`.listen('.Illuminate\\Notifications\\Events\\BroadcastNotificationCreated'`
after that change it was listening as expected to the notifications.

indigoram89:
I resolved this using namespace option.

In my case it:
`namespace: ''`

Full code:
`window.Echo = new Echo({
    namespace: ''
});`

In component:
`.listen('updated', e => {})`

Documentation:
https://laravel.com/docs/5.4/broadcasting#namespaces

matrunchyk:
Same here. Right after the update. I suspect it caused by [this commit](https://github.com/laravel/echo/commit/7b7dd5f1bd0b6bf60140ce6d2594b3fd0357eb7c).

msonowal:
this got fixed in v1.3.2

carlcassar:
[...]

Issue #137, opened 2017-07-28
Notification Name Event 


yaien:
Notification event format actually works on .listen('.Illuminate\\\Notifications\\\Events\\\BroadcastNotificationCreated', console.log);
**fix event-format.ts or change notification event to '.Illuminate\\\Notifications\\\Events\\\BroadcastNotificationCreated'**

matrunchyk:
There is an error in `src/util/event-formatter.ts`.

There shouldn't be a `return` [there](https://github.com/laravel/echo/commit/7b7dd5f1bd0b6bf60140ce6d2594b3fd0357eb7c#diff-22ec490a364e145c52e6a265271e4fd3R30)

changing this line to `event = event.substr(1);` fixes the issue

ellisio:
You should be doing the following:

```
.listen('.Illuminate.Notifications.Events.BroadcastNotificationCreated', console.log);
```

Out of curiosity, where in the documentation does it say to use the version with slashes? I can't find it.

ellisio:
You should be doing the following:

```
.listen('.Illuminate.Notifications.Events.BroadcastNotificationCreated', console.log);
```

Out of curiosity, where in the documentation does it say to use the version with slashes? I can't find it.

ellisio:
Discussion is in #139 for this issue. Looks like this can be resolved by updating the `channel.ts` call to the OP's implementation. I miss-read your OP. Sorry about that.

PR #139, opened 2017-07-31
Fixes Not receiving notifications

matrunchyk:
Fixes https://github.com/laravel/echo/issues/135 and https://github.com/laravel/echo/issues/137

### DIFF
--- a/src/util/event-formatter.ts
+++ b/src/util/event-formatter.ts
@@ -27,7 +27,7 @@ export class EventFormatter {
      */
     format(event: string): string {
         if (event.charAt(0) === '.' || event.charAt(0) === '\\') {
-            return event.substr(1);
+            event = event.substr(1);
         } else if (this.namespace) {
             event = this.namespace + '.' + event;
         }


taylorotwell:
Why does this keep breaking?

taylorotwell:
I'm so tired of going back and forth on this. Is anyone testing this with an actual application? Are we just shooting in the dark hoping shit works?

taylorotwell:
Ping @ellisio

taylorotwell:
I'm going to have to build a little demo Echo app myself today to really make sure all this shit is working.

matrunchyk:
@taylorotwell Well it should work properly with `presence`, `private` wrappers as well as with raw `listen` method. @ellisio please let us know does it work for your case as well.

taylorotwell:
Let's stop saying "it should work", etc. Either let's make sure it works or stop submitting PRs.

taylorotwell:
Ping @themsaid, we need a test application that tests every Echo situation in the documentation.

ellisio:
I'll take a look at the thread this afternoon. Busy conducting interviews this AM.

ellisio:
Just quickly looking over this code, this fix would break dot syntax. By replacing all dots with `\` you'd cause the developer to do `.listen('server\\created')` and not `.listen('server.created')`.

tlaverdure:
I'm going to take a look at this as well.

themsaid:
@taylorotwell on it

ellisio:
Okay, so I just dug into the core of Echo and these are the following patterns we need to support:

- Notifications: `.Illuminate.Notifications.Events.BroadcastNotificationCreated` -> `Illuminate\\Notifications\\Events\\BroadcastNotificationCreated`
- App\Events: `Server.Created` -> `App\\Events\\Server\\Created`
- App\Events: `Server\\Created` -> `App\\Events\\Server\\Created`
- FooBar\Events: `.FooBar.Events.MyEvent` -> `FooBar\\Events\\MyEvent`
- FooBar\Events: `\\FooBar\\Events\\MyEvent` -> `FooBar\\Events\\MyEvent`
- Dotsyntax: `.server.created` => `server.created`

I'm poking around the core to see where the disconnect is and why the above can't work. Obviously it's not just the Formatter causing issues with Notifications breaking with `1.3.1`'s implementation of the Formatter.

taylorotwell:
The format is really pretty simple. If it starts with a `.` then you just remove the dot or slash and use the name as given. Nothing else should be done to the name. `.` should not be replaced with `\` if it starts with a leading `.`. So, your "notifications" example should not be translated that way.

If the name does not start with a `.`, you prepend the namespace onto it, and all `.` in the name are replaced by `\`.

Those are the only two situations I ever intended to support and they will cover any case necessary.

taylorotwell:
I tried to summarize things as best as I can above.

ellisio:
@taylorotwell  Exactly, that is the clash and what I'm seeing when trying to dig into the core. I think the leading `\` was to allow developers to have events in different Namespaces. BUT that can be resolved by doing:

`.FooBar\\Events\\`

So in all honesty, I'm confused as to what is even broken. By this logic, when `.Illuminate.Notifications.Events.BroadcastNotificationCreated` is passed into the formatter, in `1.3.1` this will return `Illuminate.Notifications.Events.BroadcastNotificationCreated` as expected. So why are people's notifications breaking? Especially since that is how Laravel broadcasts events in that manner. 

My bet is they implemented `receivesBroadcastNotificationsOn` and that method is returning something that is causing Echo to spew on itself.

ellisio:
@taylorotwell  Exactly, that is the clash and what I'm seeing when trying to dig into the core. I think the leading `\` was to allow developers to have events in different Namespaces. BUT that can be resolved by doing:

`.FooBar\\Events\\`

So in all honesty, I'm confused as to what is even broken. By this logic, when `.Illuminate.Notifications.Events.BroadcastNotificationCreated` is passed into the formatter, in `1.3.1` this will return `Illuminate.Notifications.Events.BroadcastNotificationCreated` as expected. So why are people's notifications breaking? Especially since that is how Laravel broadcasts events in that manner. 

My bet is they implemented `receivesBroadcastNotificationsOn` and that method is returning something that is causing Echo to spew on itself.

ellisio:
@taylorotwell as expected, looking at #137, they're doing:

```
.listen('.Illuminate\\Notifications\\Events\\BroadcastNotificationCreated')
```

When it should be:

```
.listen('.Illuminate.Notifications.Events.BroadcastNotificationCreated')
```

Where in the documentation are people finding this implementation? I can't find it anywhere...

taylorotwell:
@ellisio But `Illuminate.Notifications.Events.BroadcastNotificationCreated` isn't how the event is broadcast? It is broadcast from the backend as: `Illuminate\Notifications.Events\BroadcastNotificationCreated`.

Am I correct?

taylorotwell:
From my brief reading of the EventFormatter class currently on master branch of Echo, the implementation is correct:

screen shot 2017-07-31 at 12 09 39 pm

Removes first character only if leading `.` or leading `\` and immediately returns with no other changes.

Else, prepends namespace if we have one.

Finally, translates dots to slash.

ellisio:
I honestly don't even see how that class is being broadcasted in the first place:

https://github.com/laravel/framework/blob/5.4/src/Illuminate/Notifications/Events/BroadcastNotificationCreated.php#L84-L92

That is telling me that it'll be `App.User.1`. I'm so confused as to where that `Illuminate.Notifications.Events.BroadcastNotificationCreated` is even coming from, or why it is there. Is that from a 5.0, 5.1, 5.2, or 5.3 implementation?

matrunchyk:
This is what I see in Laravel Echo Server logs as an example:
```
Channel: private-users.afa3b365-3f01-4f85-a632-e3588c11f031
Event: Illuminate\Notifications\Events\BroadcastNotificationCreated
CHANNEL private-users.afa3b365-3f01-4f85-a632-e3588c11f031
```

Not sure about Pusher.
So as I understand, notifications are broadcasted `Illuminate\Notifications\Events\BroadcastNotificationCreated`.

ellisio:
@matrunchyk How are you triggering that event in Laravel? Sorry if this seems like a basic question, just trying to fall into the rabbit hole so I can step my way back out.

ellisio:
Alright, so it's `Illuminate\Events\Dispatcher` formatting this event payload. So what happens is Pusher, Redis, etc. are all broadcasting to the `App.User.1` channel with the event `Illuminate\Notifications\Events\BroadcastNotificationCreated`.

So it looks like the fix I have in 1.3.1 is valid, and then we need to change [this method](\Notifications\Events\BroadcastNotificationCreated) to this:

```
    notification(callback: Function): Channel {
        return this.listen('.Illuminate\\Notifications\\Events\\BroadcastNotificationCreated', callback);
    }
```

matrunchyk:
I do it by `$user->notify(new PostCreated($post));`

And how I catch it on the frontend:
```
this.echo = new Echo({
    broadcaster: 'socket.io',
    key: this.config.key,
    host: `${this.config.host}:${this.config.port}`,
    encrypted: true,
});
this.echo.private('users.afa3b365-3f01-4f85-a632-e3588c11f031').notification((notification) => {
    // 
});
```

taylorotwell:
Andrew is correct that only notification method in Echo needs to change to use \
		

		
	





On Mon, Jul 31, 2017 at 12:32 PM -0500, "Andrew Ellis" <notifications@github.com> wrote:












Alright, so it's Illuminate\Events\Dispatcher formatting this event payload. So what happens is Pusher, Redis, etc. are all broadcasting to the App.User.1 channel with the event Illuminate\Notifications\Events\BroadcastNotificationCreated.


So it looks like the fix I have in 1.3.1 is valid, and then we need to change this method to this:
    notification(callback: Function): Channel {
        return this.listen('.Illuminate\\Notifications\\Events\\BroadcastNotificationCreated', callback);
    }

PR #140, opened 2017-07-31
Fixed the event signature for Notifications.

ellisio:
This will resolve the issue for `.notification()` calls using the latest formatter in 1.3.1. Thank you everyone in #137 and #139 for helping contribute to discussing this.

@taylorotwell I know we still need an application to test all these damn scenarios, but this will attach the right event signature to the callback stack for all broadcast notifications.


taylorotwell:
Published 1.3.2 on NPM.

PR #157, opened 2017-12-09
Added ability to decide if user wants x-socket-id header

ImJustToNy:
This header made a lot of mess in my code. I think it should be optional.


stephan-v:
Why is this closed, this is exactly what is needed.

PR #165, opened 2018-01-23
Fix event formatter bug

sebastiandedeyne:
When using a "root" indicator, the dots are never replaced at the end of the function.

So `.App.Foo.Bar` became `App.Foo.Bar` instead of `App\\Foo\\Bar`.

This broke our listeners at https://github.com/spatie/dashboard.spatie.be

### DIFF
--- a/src/util/event-formatter.ts
+++ b/src/util/event-formatter.ts
@@ -27,7 +27,7 @@ export class EventFormatter {
      */
     format(event: string): string {
         if (event.charAt(0) === '.' || event.charAt(0) === '\\') {
-            return event.substr(1);
+            event = event.substr(1);
         } else if (this.namespace) {
             event = this.namespace + '.' + event;
         }


taylorotwell:
Last time we made a change related to this we broke A LOT of things. That can't happen again. We need to be very sure something is actually broken because I personally tested basically every case last time we went through this and found no problems.

sebastiandedeyne:
Well it definitely broke this line. Reverting to an older Echo version fixed it

https://github.com/spatie/dashboard.spatie.be/blob/master/resources/assets/js/mixins/echo.js#L9

How about I add some unit tests to the PR so we're sure it never happens again? Would have to add a test runner to the repo though, in case you're interested, let me know which you prefer.

sebastiandedeyne:
And apparently there was a similar comment on the original PR:

https://github.com/laravel/echo/pull/134#issuecomment-318635047

taylorotwell:
Published 1.3.4.

mpyw:
@taylorotwell This PR breaks official documentation. It's not a minor update.

mpyw:
I would suggest:

- It should be published as 2.0
- Documentation should be changed:

> **If you don't need to prepend the application's namespace to the event,** you should make sure to register your listener with a leading `.` character:
> 
> ```js
> .on('.server.created', function (e) {
>     ....
> });
> ```
>
> If you customize the broadcast name using the `broadcastAs` method, you should make sure to **use `on()` instead of `listen()`**.
> 
> ```js
> .on('.server.created', function (e) {
>     ....
> });
> ```

eddiclin:
@mpyw I also met this problem today. And I read the source code, the `listen` and `on` function's return value have a little different. The `listen` function returns `this`, the `on` function returns `undefined`.

PR #166, opened 2018-01-23
Vue Resource<0.8.0 compatibility

carduz:


taylorotwell:
I don't think we need to be supporting old vue-resource. Vue-resource is not maintained anymore.

Issue #167, opened 2018-01-27
[Version 1.3.4] Can't receive events after update


kiriunin:
Version 1.3.4 is not receiving any events unlike the previous version (1.3.3).

mpyw:
We should revert this PR

[Fix event formatter bug by sebastiandedeyne · Pull Request #165 · laravel/echo](https://github.com/laravel/echo/pull/165)

mpyw:
@kiriunin Workaround

```php
/**
 * The event's broadcast name.
 *
 * @return string
 */
public function broadcastAs()
{
    return 'server.created';
}
```

```diff
- .listen('.server.created', function (e) {
+ .on('server.created', function (e) {
     ....
 });
```

kiriunin:
@mpyw Thank you, it works for me!

@taylorotwell This PR breaks official documentation...

mpyw:
Merged

PR #168, opened 2018-02-01
Revert "Fix event formatter bug"

mpyw:
This reverts commit 3780e7ff42d246341d11f69039073804a37817a7. (PR #165)

Due to: [[Version 1.3.4] Can't receive events after update · Issue #167 · laravel/echo](https://github.com/laravel/echo/issues/167)

### DIFF
--- a/src/util/event-formatter.ts
+++ b/src/util/event-formatter.ts
@@ -27,7 +27,7 @@ export class EventFormatter {
      */
     format(event: string): string {
         if (event.charAt(0) === '.' || event.charAt(0) === '\\') {
-            event = event.substr(1);
+            return event.substr(1);
         } else if (this.namespace) {
             event = this.namespace + '.' + event;
         }


taylorotwell:
OK. I am *sick* of going back and forth on this. What is happening here and how is it fixed for good?

taylorotwell:
@sebastiandedeyne

taylorotwell:
I agree the previous behavior was actually correct.

morloderex:
We need to add some tests for this class..

taylorotwell:
Yep. It's getting ridiculous. I was pretty certain I had at least manually confirmed it's all correct by hand last time. I'm unfortunately not familiar enough with JS testing to write tests for this. If someone is please contribute them. The behavior as it stands now is correct.

taylorotwell:
The `.` is a literal indicator and anything after that is returned unchanged since someone people do events like `foo.bar`.

sebastiandedeyne:
As I've states originally in the PR, I'd be willing to add some tests here. Is there any framework you prefer? I generally use Jest but anything's possible really

mpyw:
Also we can improve readability by changing:

```ts
    /**
     * Format the given event name.
     *
     * @param  {string} event
     * @return {string}
     */
    format(event: string): string {
        if (event.charAt(0) === '.' || event.charAt(0) === '\\') {
            return event.substr(1);
        } else if (this.namespace) {
            event = this.namespace + '.' + event;
        }

        return event.replace(/\./g, '\\');
    }
```

into:

```ts
    /**
     * Format the given event name.
     *
     * @param  {string} event
     * @return {string}
     */
    format(event: string): string {
        if (event.charAt(0) === '.' || event.charAt(0) === '\\') {
            return event.substr(1);
        }
        if (this.namespace) {
            event = this.namespace + '.' + event;
        }
        return event.replace(/\./g, '\\');
    }
```

It's useless `elesif`.

sebastiandedeyne:
> The . is a literal indicator and anything after that is returned unchanged since sometimes people do events like foo.bar.

Fwiw, I expect `.` to mean "root namespace" (that was also the idea behind the original PR #11)

mpyw:
@sebastiandedeyne Actually I agree with you.

True solution:
https://github.com/laravel/echo/pull/165#issuecomment-361493790

However I also agree those method names (`listen` and `on`) are ambiguous...

mpyw:
Simply because it broke official documentation example.

PR #207, opened 2018-10-31
[WIP] General cleanup, and change export of main Echo module

FrozenDroid:
May or may not be breaking. I'll need some help with testing this since I exclusively use this module with TypeScript and Webpack.

The export has been changed to work nicer with TypeScript, and should work for anything, really.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,14 +27,16 @@
   "devDependencies": {
     "@types/jest": "^22.1.0",
     "@types/node": "^6.0.85",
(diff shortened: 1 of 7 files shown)


FrozenDroid:
Oh well, that's why I added a WIP label :P

driesvints:
@FrozenDroid yeah, it's best to only send in finished PRs since otherwise they'll just stack up and we'll lose our overview of them.

Issue #211, opened 2018-12-01
SyntaxError: Unexpected token export (1.5.0)


udanpe:

driesvints:
Please open a new issue with enough info instead of just posting an empty issue.

udanpe:
@driesvints This error don't need description. And you know about this bug. But how say @tlaverdure if you don't use module as intended (with mix and pusher) your are not a people and must clone repository for self. React create app, react native, cordova,  nextjs, nuxtjs and other all must suffer. This is not serious.

driesvints:
@udanpe I have no idea how this bug could be reproduced. What steps do you take?

udanpe:
@driesvints https://github.com/laravel/echo/pull/210 I use next.js with other modules. I think 100-150 and all this modules convert es6 to commonjs. You can show me just one public module who dont do this?

driesvints:
You're still not posting any steps. Please share your code where this is breaking things and post steps you take how you install the package.

udanpe:
@driesvints it's not my fault) just convert es6 to commonjs. Last node lts 10.14 don't support es6 import export

udanpe:
@driesvints 
module.exports = Echo // support
export default Echo // don't support

udanpe:
@driesvints But if you want step by step, ok. Install next.js https://github.com/zeit/next.js/ or another frontend framework import Echo run in dev mode and you can see this error. Commonjs it's standard.

driesvints:
@udanpe we can't accommodate for every framework out there. It's either the current version or another one. 

Anyway, my knowledge about JS is too limited to provide a proper answer for this. Maybe @tlaverdure who sent in the PR could answer this better.

samuelterra22:
I am using Nuxt.js and received the same error after upgrading to version 1.5.0

![image](https://user-images.githubusercontent.com/11915449/49382315-46377b00-f6fd-11e8-8a9c-8eeed169aa7b.png)

udanpe:
@driesvints maybe you have @tlaverdure contacts? It is very and very not professional. @taylorotwell what your thing about this?

udanpe:
@driesvints maybe you have @tlaverdure contacts? It is very and very not professional. @taylorotwell what you think about this?

udanpe:
If your really want use es6 you must create other file echo.common.js for other people

tlaverdure:
Sure, I'll send a PR to also package a echo.common.js file. No need to @ me so many times...

PR #212, opened 2018-12-03
Add commonjs output to distribution.

tlaverdure:
Some developers are using this package server side, without a bundler like Webpack. This includes a CommonJs output as the [main](https://docs.npmjs.com/files/package.json#main) module for those cases, and ES6 exports should still be discovered using the `module` property of the package.json file.

Both should work depending on the build tools the developer is using. For example, Rollup and Webpack use this field to determine which is appropriate.

```js
import Echo from "laravel-echo"
```

```js
const Echo = require("laravel-echo");
```

Addresses: https://github.com/laravel/echo/issues/211

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,8 @@
   "name": "laravel-echo",
   "version": "1.4.2",
   "description": "Laravel Echo library for beautiful Pusher and Socket.IO integration",
-  "main": "dist/echo.js",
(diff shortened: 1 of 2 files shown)


tlaverdure:
It gets compiled at build time and placed in the dist folder which is published.

Build - npm run build

Publish - npm run publish

PR #214, opened 2018-12-05
Add iife output to dist

han003:
Add iife file to dist that developers can use in a `script` tag.


taylorotwell:
What does this do exactly?

han003:
Hi.
When I tried using laravel echo in my project I could not get it to work by
including it in a script tag.
It seems it can only be used with a require or import statement, which I
unfortunately can't use.
Because that requires me to use something like webpack, and I don't have
that. I have a basic index.html file where scrips are included via script
tags.
This iife file creates a file where Echo is global and can be included into
projects by using script tags.

Issue #239, opened 2019-08-06
Typescript: listen property does not exist on type 'Presence Channel'


willypuzzle:
I use laravel echo with Typescript and I have a problem on listen property, I receive a 

> TS2339: Property 'listen' does not exist on type 'PresenceChannel'.

In this segment of code:

```
    echo.join(channel)
            .here((users) => {
                EventBusMain.$emit(EVENTS.HERE_MAIN_CHANNEL_EVENT, {users});
            })
            .joining((user) => {
                EventBusMain.$emit(EVENTS.JOIN_MAIN_CHANNEL_EVENT, {user});
            })
            .leaving((user) => {
                EventBusMain.$emit(EVENTS.LEAVE_MAIN_CHANNEL_EVENT, {user});
            })
            .listen('.event', (e) => {
                EventBusMain.$emit(e.event, e.data)
            });
```

The version of laravel echo I'm using is 1.5.4

driesvints:
Hi there,

Thanks for reporting but it looks like this is a question which can be asked on a support channel. Please only use this issue tracker for reporting bugs with the library itself. If you have a question on how to use functionality provided by this repo you can try one of the following channels:

- [Laracasts Forums](https://laracasts.com/discuss)
- [Laravel.io Forums](https://laravel.io/forum)
- [StackOverflow](https://stackoverflow.com/questions/tagged/laravel)
- [Discord](https://discordapp.com/invite/KxwQuKb)
- [Larachat](https://larachat.co)
- [IRC](https://webchat.freenode.net/?nick=laravelnewbie&channels=%23laravel&prompt=1)

However, this issue will not be locked and everyone is still free to discuss solutions to your problem!

Thanks.

willypuzzle:
I think this is a problem of Echo interface bug of typescript because that property is declared in the [documentation](https://laravel.com/docs/5.8/broadcasting#broadcasting-to-presence-channels) but not declared in the typescript interface. I think this is a bug of typescript.

driesvints:
@willypuzzle I actually closed it because I didn't saw an issue template. If you're reporting bugs please fill it out next time.

I indeed don't see the `listen` method on the PresenceChannel interface but it is defined as an abstract method. So not sure why Typescript is complaining. If the return type has it defined. Maybe I don't understand Typescript well enough. Which connector are you using?

willypuzzle:
I'm using typescript version 2.9.2

driesvints:
I meant pusher or socket.io

[...]

PR #241, opened 2019-08-21
Added listen methods to presence channels and interface - prevent Angular build errors

Mathem8tic:
@driesvints this is most likely not the correct solution, but it does prevent the errors which are stopping the build from succeeding when using Angular CLI.  You will see I just added the listen method to the various PresenceChannels.

I hope this helps!

### DIFF
--- a/src/channel/null-presence-channel.ts
+++ b/src/channel/null-presence-channel.ts
@@ -32,4 +32,12 @@ export class NullPresenceChannel extends NullChannel implements PresenceChannel 
     whisper(eventName: string, data: any): NullPresenceChannel {
         return this;
     }
+
+    /**
+     * Listen for an event on the channel instance.
+     */
+    listen(event: string, callback: Function): NullPresenceChannel {
+        return this;
+    }
+
 }
--- a/src/channel/null-presence-channel.ts
+++ b/src/channel/null-presence-channel.ts
@@ -39,5 +39,4 @@ export class NullPresenceChannel extends NullChannel implements PresenceChannel 
     listen(event: string, callback: Function): NullPresenceChannel {
         return this;
     }
-
 }
--- a/src/channel/presence-channel.ts
+++ b/src/channel/presence-channel.ts
@@ -1,3 +1,4 @@
+
 /**
  * This interface represents a presence channel.
  */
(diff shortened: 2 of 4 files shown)


taylorotwell:
If it's most likely not the correct solution let's not do it. 😬

Mathem8tic:
Sorry, it works.  I just don't know if it is as clean as your standards.  I wanted @driesvints to take a look since he requested a PR.

Cheers.

driesvints:
I need to be able to reproduce the issue first. Please provide a repo that reproduces it.

driesvints:
@Mathem8tic ah just saw in the issue that you did. I'll try to have a look soon.

sebastiandedeyne:
@Mathem8tic Why don't you consider this to be the right solution?

Mathem8tic:
I think it might be correct.  I just assumed those methods were left out
for a reason that I was unaware of.  I can confirm that the solution works
nicely for my project.

Feel free to pull it in if you are satisfied with the additions.

Take care.

driesvints:
Okay, re-opening then.

taylorotwell:
Can someone let me know what issue we are attempting to fix?

Mathem8tic:
#239

taylorotwell:
This doesn't maky any sense. Pusher presence channel *ALREADY* extends PusherChannel which already has a `listen` method with an implementation. Surely overriding that with a new `listen` method which just does `return this` doesn't seem like the best idea and will break applications?

driesvints:
I really don't know sorry. I can reproduce this but I have too limited knowledge of typescript or javascript to know what the best solution is.

mengyang1229:
@Mathem8tic @taylorotwell @driesvints @sebastiandedeyne @ellisio 

Here is my solution and hope it can help you. Thank you.

[https://github.com/laravel/echo/issues/239#issuecomment-550114637](https://github.com/laravel/echo/issues/239#issuecomment-550114637)

johnpaulmedina:
So I found that there isn't a bug or an issue with the interface at all with the PresenceChannel.

Just assign as the type for your LaravelEchoChannel ng-cli wont throw any errors.
```es6
  private echo: Echo;
  private LaravelEchoChannel: any;

  LaravelEchoChannel = echo.join(channel)
     .here((users: any) => {
       log.debug('[Presence][Channel]', users);
     })
     .joining((user: any) => {
       log.debug('joining', user.name);
     })
     .leaving((user: any) => {
       log.debug('leaving', user.name);
     });

  LaravelEchoChannel
    .listen('.event', (event: string) => {
       log.debug('event');
    });
```

This seems more of a ng-cli ts config issue.

PR #242, opened 2019-09-06
[1.x] Try to get CSRF token from cookie

m1guelpf:
The Laravel skeleton was recently updated to stop explicitly setting the CSRF token on the axios client, as it can automatically get it from the XSRF-TOKEN Laravel adds by default. This PR introduces the same behavior on Echo.

### DIFF
--- a/src/connector/connector.ts
+++ b/src/connector/connector.ts
@@ -50,9 +50,9 @@ export abstract class Connector {
         let cookie;
 
         const readCookie = function(name) {
-          var match = document.cookie.match(new RegExp('(^|;\\s*)(' + name + ')=([^;]*)'));
-          return (match ? decodeURIComponent(match[3]) : null);
-        },
+            var match = document.cookie.match(new RegExp('(^|;\\s*)(' + name + ')=([^;]*)'));
+            return match ? decodeURIComponent(match[3]) : null;
+        };
 
         if (typeof window !== 'undefined' && window['Laravel'] && window['Laravel'].csrfToken) {
             return window['Laravel'].csrfToken;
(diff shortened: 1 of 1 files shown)


taylorotwell:
Does this work since content of that cookie is an **encrypted** version of the CSRF token, not the token itself?

m1guelpf:
As you can see, Laravel decrypts the header here: https://github.com/laravel/framework/blob/6.x/src/Illuminate/Foundation/Http/Middleware/VerifyCsrfToken.php#L153

The only thing needed to make this work would be to change the header name from X-CSRF to X-XSRF when sending the encrypted versiom. Will PR that later today

taylorotwell:
Honestly I'm fine not messing with this.

mpskovvang:
I was about to submit a PR with a slightly different approach to setting the `X-XSRF-TOKEN` header, but realized it won't work with a SPA as the header can't be changed after constructing Pusher.

I came up with this solution, which ensures the header is always kept up-to-date:

```JavaScript
const cookie = function (name) {
    let match = document.cookie.match(new RegExp('(^|;\\s*)(' + name + ')=([^;]*)'));
    return (match ? decodeURIComponent(match[3]) : null);
}

const xsrf = (urlmatch) => {
    let open = XMLHttpRequest.prototype.open;
    XMLHttpRequest.prototype.open = function (method, url) {
        open.apply(this, arguments);

        if ((new URL(url, window.location)).toString().includes(urlmatch)) {
            this.setRequestHeader('X-XSRF-TOKEN', cookie('XSRF-TOKEN'))
        }
    }
};

xsrf('https://example.org/broadcasting/auth')

new Echo(...)
```

PR #245, opened 2019-09-13
Fixes #244

### LINKED ISSUE #244: Use jQuery.ajaxFilter instead of jQuery.ajaxSetup
`jQuery.ajaxSetup` will be easily get overridden when we are configuring `jQuery.ajax`, it is better to use `jQuery.ajaxPrefilter` which is like a middleware. same as `axios`.

Currently it is like this:
```js
$.ajaxSetup({
  beforeSend (xhr) {
    // We are adding socketId header here
  }
});

$.ajax({
  beforeSend (xhr) {
    // I want to do something else here, but the socketId header will be removed
  }
});
```
It should be like this: 
```js
$.ajaxPrefilter(function (options, originalOptions, xhr) {
  // We are adding socketId header here
});

$.ajax({
  beforeSend (xhr) {
    // I want to do something else here, the socketId will NOT be removed
  }
});
```

sharifzadesina:


taylorotwell:
Please provide a full issue description.

sharifzadesina:
@taylorotwell There is already an issue for this PR, should I write them here again?

driesvints:
@sharifzadesina make sure you don't miss the PR template when submitting prs. Always add an explanation to your pr and not just link to an issue. It's hard for taylor to always parse issues for every pr he reviews (there's a lot each day)

PR #248, opened 2019-09-14
feat: add option `disableAutoRegisterInterceptors`

### LINKED ISSUE #97: If this fails so does your app, it should be able to fail and continue
Using ember I made a service for echo and had made Stubs for when pusher isn't configured or is failing because the app should run with or without pusher.

This worked to stop pusher down causing our whole app to be down until the jQuery `beforeSend` header was set. Now I have to hack echo, extend it to stop all these things it registers before even knowing if it can get a valid connection.

You should be able to disable any of these registrations https://github.com/laravel/echo/blob/master/src/echo.ts#L32. Or even better explicitly specify what you are using and only register that.

### LINKED ISSUE #152: x-socket-id header default causing issues.
Why is the `X-Socket-Id` set per default?

This causes a shitload of problems with most API's and it would be more beneficial to add it whenever you need to call a socket endpoint, rather than to remove it when it causes problems.

The Github API for example does not allow it in their preflight response. For Axios I have to do this all over the place now:

```
axios.get('user/repos', {
    transformRequest: [function (data, headers) {
        delete headers['X-Socket-Id'];
        return data;
    }]
})
```

ricardogobbosouza:
Resolve #152 
Resolve #97

### DIFF
--- a/src/echo.ts
+++ b/src/echo.ts
@@ -21,7 +21,10 @@ export default class Echo {
     constructor(options: any) {
         this.options = options;
         this.connect();
-        this.registerInterceptors();
+
+        if (!this.options.disableAutoRegisterInterceptors) {
+            this.registerInterceptors();
+        }
     }
 
     /**
--- a/src/echo.ts
+++ b/src/echo.ts
@@ -22,7 +22,7 @@ export default class Echo {
         this.options = options;
         this.connect();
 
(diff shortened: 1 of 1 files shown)


AUnhurian:
The name of the option has been changed to: `withoutInterceptors` in later versions

PR #249, opened 2019-09-14
Update module

### LINKED ISSUE #217: Is there a changelog?
So, I looked through the repo, looked through the official laravel docs, and anywhere else I could think of, but there doesn't seem to be a changelog.

Am I missing it somewhere?

ricardogobbosouza:
- Update dependencies
- Sort package.json
- Split jest config into a file `jest.config.js`
- Add script `release` to generate version with `standard-version` (Resolve #217)


driesvints:
@ricardogobbosouza heya. Thanks for this. We don't use yarn however so I removed its lock file again. I also removed the package-lock.json file because we don't commit it. And I've reverted the changes to the gitignore because these are mostly things you should keep in your global gitignore.

PR #251, opened 2019-09-22
Update connector.ts

subotkevic:
It turns out that React Native has `document` defined, it contains some properties, but it doesn't have a `querySelector` property.

This is a quick, additional, check to handle that.


subotkevic:
When are you planning to draft a new release including this fix? Thanks.

driesvints:
@subotkevic today is release day so today probably

subotkevic:
Thanks for the info, @driesvints!

PR #256, opened 2019-10-08
Added some prefix helpers

webdevium:
- Added basic prefix helpers for channel name: _privatify_ and _presencify_


taylorotwell:
Unclear what bug these are fixing?

Issue #263, opened 2020-04-02
Support pusher end-to-end encryption


it-can:
Would be nice to have support for end-to-end encryption

https://github.com/pusher/pusher-http-php#end-to-end-encryption-beta

driesvints:
So this is just to get it working client side right? Because you can already add the options server-side in your config.

it-can:
yeah this seems to be possible within laravel, 

https://github.com/laravel/framework/pull/31559

But I think we also need an option to subsribe  to  ```private-encrypted-``` channels... this is not possible in echo as far as I know now...

driesvints:
I'll mark this as an enhancement for now. Feel free to pr.

it-can:
ok will try to find some time to create a PR...

it-can:
@driesvints ok added a PR, can you take a look if I did it correct?

derekmd:
@driesvints: I believe this issue can be closed? "laravel/echo" 1.7.0 and "laravel/framework" 6.17.0 changes allow private encrypted channels.

PR #264, opened 2020-04-02
add pusher private-encrypted #263

it-can:
Ok added support for private-encrypted channels (#263) (only for Pusher). Hope this is correct.. I don't use Ts a lot...

### DIFF
--- a/src/channel/index.ts
+++ b/src/channel/index.ts
@@ -2,7 +2,7 @@ export * from './channel';
 export * from './presence-channel';
 export * from './pusher-channel';
 export * from './pusher-private-channel';
-export * from './pusher-private-encrypted-channel';
+export * from './pusher-encrypted-private-channel';
 export * from './pusher-presence-channel';
 export * from './socketio-channel';
 export * from './socketio-private-channel';
--- a/src/channel/index.ts
+++ b/src/channel/index.ts
@@ -2,6 +2,7 @@ export * from './channel';
 export * from './presence-channel';
 export * from './pusher-channel';
 export * from './pusher-private-channel';
+export * from './pusher-private-encrypted-channel';
 export * from './pusher-presence-channel';
 export * from './socketio-channel';
 export * from './socketio-private-channel';
--- a/src/channel/pusher-encrypted-private-channel.ts
+++ b/src/channel/pusher-encrypted-private-channel.ts
@@ -3,11 +3,11 @@ import { PusherChannel } from './pusher-channel';
 /**
  * This class represents a Pusher private channel.
  */
-export class PusherPrivateEncryptedChannel extends PusherChannel {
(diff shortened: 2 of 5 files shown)


driesvints:
Seems to be good.

@taylorotwell note that this is to support https://github.com/laravel/framework/pull/31559

taylorotwell:
@it-can did you actually build this and test it using `npm link`?

it-can:
@taylorotwell yes it just tested this. works perfectly.

In Laravel set in the config/broadcasting.php file, a base64 encoded masterkey.

```
'pusher' => [
            'driver'  => 'pusher',
            'key'     => env('PUSHER_APP_KEY'),
            'secret'  => env('PUSHER_APP_SECRET'),
            'app_id'  => env('PUSHER_APP_ID'),
            'options' => [
                        'useTLS'       => true,
                        'cluster'      => 'eu',
                        'encryption_master_key_base64' => 'mysecretbase64encodedkey',
            ],
        ],
```

Broadcast to a EncryptedPrivateChannel channel.

```
namespace App\Events;

use Illuminate\Broadcasting\Channel;
use Illuminate\Queue\SerializesModels;
use Illuminate\Broadcasting\InteractsWithSockets;
use Illuminate\Broadcasting\EncryptedPrivateChannel;
use Illuminate\Contracts\Broadcasting\ShouldBroadcast;

class CalenderRefresh implements ShouldBroadcast
{
    use InteractsWithSockets, SerializesModels;

    /**
     * Create a new event instance.
     *
     * @return void
     */
    public function __construct()
    {
        //
    }

    /**
     * Get the channels the event should broadcast on.
     *
     * @return Channel|array
     */
    public function broadcastOn()
    {
        return new EncryptedPrivateChannel('calendar');
    }
}
```

Subscribe with Echo to a privateEncrypted channel
```
Echo.privateEncrypted('calendar').listen('CalenderRefresh', function () {
            reloadCalendar();
        });
```

This works for me

taylorotwell:
One thing I notice is a mismatch between backend and frontend. On backend we say "EncryptedPrivate" but on this we say "privateEncrypted"... I would change it to "encryptedPrivate" here in Echo so it matches.

it-can:
> One thing I notice is a mismatch between backend and frontend. On backend we say "EncryptedPrivate" but on this we say "privateEncrypted"... I would change it to "encryptedPrivate" here in Echo so it matches.
 
yeah I also noticed it... Pusher uses private-encrypted... But I guess changing the naming in Laravel is a breaking change... I'll change my PR

it-can:
@taylorotwell ok renamed it

taylorotwell:
Released Echo 1.7.0 on NPM.

it-can:
Thanks Taylor!

PR #267, opened 2020-04-27
Header keys update

BKorenek:
Currently troubleshooting an issue that I thought this may be the cause of, doesn't look like it is but I wanted to do a PR to be sure no one else goes down the same rabbit hole in the future. 

I had thought the headers being mis-capitalized was causing the `toOthers` method on the broadcasted event to not be able to locate the Socket ID. However after updating I'm still having that issue. 

Alternatively the header that Laravel is looking for in the BroadcastManager to `X-Socket-Id` and Vue updated here since these header keys get normalized anyhow.

### DIFF
--- a/src/echo.ts
+++ b/src/echo.ts
@@ -106,8 +106,8 @@ export default class Echo {
     }
 
     /**
-     * Register 3rd party request interceptiors. These are used to automatically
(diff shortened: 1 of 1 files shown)


taylorotwell:
If we're not fixing anything here let's not change it.

Issue #266, opened 2020-04-16
1.7.0 release does not work in IE 11


bengallienne:
- Echo Version: 1.7.0
- Laravel Version: 6.18.6
- PHP Version: 7.3
- NPM Version: 6.14.4
- Node Version: 10.20.1

### Description:
When compiling a Laravel environment that includes the normal echo code in the app.js file, for example;

`
import Echo from 'laravel-echo';

window.Pusher = require('pusher-js');

window.Echo = new Echo({
	broadcaster: 'pusher',
	key: process.env.MIX_PUSHER_APP_KEY,
	cluster: process.env.MIX_PUSHER_APP_CLUSTER,
	encrypted: true
});
`

After compilation, viewing the page using IE11 results in a syntax error which points to this piece of code;

`class Connector {
    /**
     * Create a new class instance.
     */
    constructor(options) {
        /**
         * Default connector options.
         */
        this._defaultOptions = {
            auth: {
                headers: {},
            },
            authEndpoint: '/broadcasting/auth',
            broadcaster: 'pusher',
            csrfToken: null,
            host: null,
            key: null,
            namespace: 'App.Events',
        };
        this.setOptions(options);
        this.connect();
    }
    /**
     * Merge the custom options with the defaults.
     */
    setOptions(options) {
        this.options = Object.assign(this._defaultOptions, options);
        if (this.csrfToken()) {
            this.options.auth.headers['X-CSRF-TOKEN'] = this.csrfToken();
        }
        return options;
    }
    /**
     * Extract the CSRF token from the page.
     */
    csrfToken() {
        let selector;
        if (typeof window !== 'undefined' && window['Laravel'] && window['Laravel'].csrfToken) {
            return window['Laravel'].csrfToken;
        }
        else if (this.options.csrfToken) {
            return this.options.csrfToken;
        }
        else if (typeof document !== 'undefined' &&
            typeof document.querySelector === 'function' &&
            (selector = document.querySelector('meta[name="csrf-token"]'))) {
            return selector.getAttribute('content');
        }
        return null;
    }
}`

### Steps To Reproduce:
Build a basic laravel app with vue front end using the versions above.
Add in the code above to include Echo in something like app.js
Try to load the page in IE, it does not load
Repeat the same process using version 1.6.1 of Echo, syntax error goes away and page loads

taylorotwell:
Hmm, are you able to help us fix it?

bengallienne:
[...]

PR #270, opened 2020-05-15
IE11 fix for dist/echo.js not being transpiled to ES5

derekmd:
This covers https://github.com/laravel/echo/issues/266 to fix Laravel Echo no longer working in IE11 and other browsers that don't support native ES6.

### TLDR Fix

Add `rollup.config.js` option `extensions: ['.ts']` to explicitly whitelist the Babel plugin to handle the .ts -> ES6 transformed file.

### Should a package-lock.json be added?

Rollup was skipping plugin step `babel()` so the resulting distribution was in ES6 syntax rather than ES5.

Sometime after Laravel Echo v1.6.1 ( https://github.com/laravel/echo/commit/1cbd05381f584b4bbf161b1eed6e16753d6e6ba1 ) was tagged on Oct 1, 2019, an NPM _child_ dependency introduced a breaking change to the Rollup plugin for Babel. The step was being silently skipped so the v1.7.0-tagged dist/echo.js now being imported by Laravel apps was the plain Typescript -> ES6 transpile.

```
git reset 1cbd05381f584b4bbf161b1eed6e16753d6e6ba1
git clean -fd
rm -rf node_modules/
npm install
```

^ Using the 1.6.1 release commit, `npm run build` today generates a ES6 file, not ES5. Some nested  package.json dependency (one of over 600 packages) made a change but I still haven't figured out which. Reverting to an earlier `rollup-plugin-babel` version, still only ES6 is generated by `npm run build`.

I'd **strongly suggest introducing a package-lock.json** to prevent this from happening again.

### Why I had to update most NPM dev dependencies

1. NPM package `rollup-plugin-babel` won't allow defining that Rollup plugin option `extensions: ['.ts']`. Attempting it generates error "Plugin/Preset files are not allowed to export objects, only functions." This is caused by the old Babel 6 plugin ecosystem being used.
2. Bumping to Babel 7 requires changing the `rollup.config.js` plugin setup.
   * Preset `es2015-rollup` is replaced `@babel/preset-env` to target ES5
   * Preset `stage-2` is replaced by a handful of individual `@babel/plugin-` packages, as suggested by https://www.npmjs.com/package/@babel/preset-stage-2
3. To make this more complicated, Rollup has also changed NPM package namespacing similar to Babel. So I had to update `rollup` and change to the `@rollup/plugin-*` packages.
   * that requires bumping Echo dev environments to at least Node 10
4. However instead of upgrading `rollup-plugin-typescript` to `@rollup/plugin-typescript`, I had to replace it with fork `rollup-plugin-typescript2`.
   * `@rollup/plugin-typescript` won't transpile echo.ts to ES6 as it generates an error parsing Typescript despite the `tsconfig.json` and `typings/index.d.ts` configurations.
   * There is no upgrade path documented and Echo meets minimum requirements so I don't know what is happening. https://github.com/rollup/plugins/tree/master/packages/typescript

I need a drink.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,21 +28,25 @@
     "test": "jest"
   },
   "devDependencies": {
+    "@babel/plugin-proposal-decorators": "^7.8.3",
+    "@babel/plugin-proposal-export-namespace-from": "^7.8.3",
+    "@babel/plugin-proposal-function-sent": "^7.8.3",
+    "@babel/plugin-proposal-numeric-separator": "^7.8.3",
+    "@babel/plugin-proposal-throw-expressions": "^7.8.3",
+    "@babel/plugin-transform-object-assign": "^7.8.3",
+    "@babel/preset-env": "^7.9.6",
+    "@rollup/plugin-babel": "^5.0.0",
     "@types/jest": "^24.0.18",
     "@types/node": "^12.7.5",
-    "babel-plugin-transform-object-assign": "^6.22.0",
-    "babel-preset-es2015-rollup": "^3.0.0",
-    "babel-preset-stage-2": "^6.24.1",
     "jest": "^24.9.0",
-    "rollup": "^1.21.2",
-    "rollup-plugin-babel": "^4.3.3",
-    "rollup-plugin-typescript": "^1.0.1",
+    "rollup": "^2.10.2",
+    "rollup-plugin-typescript2": "^0.27.1",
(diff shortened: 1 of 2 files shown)


taylorotwell:
Thanks. I have published 1.8.0.

driesvints:
Thanks @derekmd!

PR #277, opened 2020-07-22
Run build on CI

kamui545:
Are there any reasons that only tests are run on the CI?
They are very limited and I think `npm run build` should be run as well to ensure that it is at least compiling properly.


driesvints:
Good idea. Thanks!

Issue #177, opened 2018-03-23
Support "pusher:subscription_error" event


tillkruss:
Support for `pusher:subscription_error` would be nice. It only applied to `private` and `presence` channels.

`pusher:subscription_succeeded` could be moved to basic pusher channel, because it's triggered for all channel types.

https://pusher.com/docs/client_api_guide/client_events#subscription_error

driesvints:
How would this exactly work? Something like an `.error()` method? Can't you simply catch these errors in your code?

charleswolfe:
I could see subscription errors coming back in firefoxes console. I wanted to be able to execute a function when the subscription failed. 

I prefer something like 'subscription_error', however 'error' would be fine if there was a type for 'subscription_error', so I handle that very specific case.

Also thanks for looking into this @driesvints , knowing that we didn't make the socket connection and being able to handle it allows us to write less code. For example: dealing with a user that logs in to an application simultaneously using multiple browsers.

tillkruss:
@driesvints: I guess I'd like to handle errors in a callback and if there is no callback maybe rethrow the exception or just call `console.error()`.

```js
Echo.join(`chat.${roomId}`)
    .here(...)
    .joining(...)
    .leaving(...)
    .listen(...)
    .error(error => {
        //
    });
```

driesvints:
Is there a way to do a type check on the error? Something like this:

```js
Echo.join(`chat.${roomId}`)
    .here(...)
    .joining(...)
    .leaving(...)
    .listen('NewMessage', ...)
    .error(err => {
        if (err.type === 'subscription') {
             // do whatever type related
        }
        
        // Sentry.captureException(err);
    });
```

Reason why I want to do it this way is to just provide a generic way of catching errors so people can do whatever they want.

tillkruss:
That'd be great.

```js
Echo.join(`chat.${roomId}`)
    .here(...)
    .joining(...)
    .leaving(...)
    .listen('NewMessage', ...)
    .error((err, type) => {
        // type = subscription
        // err = Error object
    });
```

driesvints:
Okay so I don't know when I'll have time to work on this but feel free to send in a PR if you get to it before me.

driesvints:
Sent a PR for this here: https://github.com/laravel/echo/pull/284

@tillkruss would be great if you could give feedback on my notes on the private channel thing on the PR.

tillkruss:
@driesvints: LGTM. Haven't tested it, I don't think we use that code anymore.

PR #284, opened 2020-07-31
Implement error handling

driesvints:
This PR implements support for Pusher's `pusher:subscription_error` event through a dedicated and generic `error` callback handler. It'll pass along the status as its first argument when a subscription error occurs.

I couldn't find any proper docs on Socket.io's events so I wouldn't know what the equivalent is for Socket.io.

~@tillkruss noted that this is applicable for Private Channel as well. I noticed `here` is also only applied on `PresenceChannel`. I feel like like both `here` and `error` and maybe even `joining`, `leaving` and `whisper` should too? Maybe `PrivateChannel` should extend `PresenceChannel`? What are your thoughts @taylorotwell?~

Moving just `.error` to the abstract `PusherChannel` class.

Because this adds a new method on an interface this probably needs a new major release. Although I don't know what the policy is for Echo so this might just be a minor?

Resolves https://github.com/laravel/echo/issues/177

### DIFF
--- a/src/channel/channel.ts
+++ b/src/channel/channel.ts
@@ -32,7 +32,7 @@ export abstract class Channel {
     abstract stopListening(event: string): Channel;
 
     /**
-     * Stop listening for a whispser event on the channel instance.
+     * Stop listening for a whisper event on the channel instance.
      */
     stopListeningForWhisper(event: string): Channel {
         return this.stopListening('.client-' + event);
--- a/src/channel/channel.ts
+++ b/src/channel/channel.ts
@@ -37,4 +37,9 @@ export abstract class Channel {
     stopListeningForWhisper(event: string): Channel {
         return this.stopListening('.client-' + event);
     }
+
+    /**
+     * Register a callback to be called anytime an error occurs.
+     */
+    abstract error(callback: Function): Channel;
 }
--- a/src/channel/null-channel.ts
+++ b/src/channel/null-channel.ts
@@ -33,6 +33,13 @@ export class NullChannel extends Channel {
     }
 
     /**
+     * Register a callback to be called anytime an error occurs.
+     */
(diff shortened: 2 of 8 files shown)


driesvints:
Pushed some changes that just allow for an error callback on both present and private channels. By default pusher has the error handling of `pusher:subscription_error` on both of these.

xbrunosousa:
Hi! Is it possible to use it in the echo constructor? 
```js
(new Echo ({...})
```

driesvints:
Not with this implementation.

Patabugen:
Hey!

I'm trying to implement this to detect two scenarios (and also just basic error handling) but struggling to figure out the syntax.

I don't fully understand the old comment which has been struck-out, it seems to have said that this works on Private, but does it being crossed out mean it doesn't?

The scenarios I'm trying to catch:
a) The laravel-echo-server is not running (e.g. invalid hostname)
b) `subscription_error` specifically (but also I'd like to catch all errors)

Would someone mind adding error-handling documentation to the Echo docs? Or if you can chuck some examples at me I can make a PR.

*This is not valid code* but these are some of the things I've tried over the past hour or two from reading various bits of source/PR messages/stack overflow questions/etc.

The only bit which seems to work is `.listen('subscription_error', ...`, but 

```
    window.Echo = new Echo({
        broadcaster: 'socket.io',
        host: process.env.MIX_WEBSOCKET_HOST,
    }).error(function(error, type) {
           console.error(error);
           console.error(type);
    }).listen('subscription_error', (e) => {
    console.log
   });


    Echo.private(`my_private_channel`)
        .on('subscription_error', function (status, error) {
            console.log(status);
            console.log(error);
        })
        .listen('subscription_error', (e) => {
            console.log(e);
        })
        .listen('DriveStarted', (e) => {
    })
```

PR #291, opened 2020-12-16
Add optional callback argument to stopListening()

larsnystrom:
This makes it possible to unregister a single callback for an event, without unregistering all other callbacks for that event on that channel. This is very helpful for example when using react hooks, where you need to stop listening as the component is unmounted. Here's an example component using this new functionality:

    const MyEventMessagesList = () => {
        const [messages, setMessages] = useState([]);

        useEffect(() => {
            const pushMessage = message => setMessages(prev => [...prev, message]);
            const channel = Echo.private("my.channel")
                .listen("MyEvent", pushMessage);

            // It used to be cumbersome to unregister a single callback
            // for an event, but now you can just do this:
            return () => channel.stopListening("MyEvent", pushMessage);
        }, [setMessages]);

        return (
            <ul>
                {messages.map((m, i) =>
                    <li key={i}>{m}</li>
                )}
            </ul>
        )
    }

### DIFF
--- a/src/channel/channel.ts
+++ b/src/channel/channel.ts
@@ -29,13 +29,13 @@ export abstract class Channel {
     /**
      * Stop listening to an event on the channel instance.
      */
-    abstract stopListening(event: string): Channel;
+    abstract stopListening(event: string, callback?: Function): Channel;
 
     /**
      * Stop listening for a whisper event on the channel instance.
      */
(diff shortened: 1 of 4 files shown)


larsnystrom:
Sorry, I just realised this won't work in the socket.io channel. I'm going to have to have another go at this. Closing for now.

Issue #294, opened 2020-12-30
Make .whisper and .listenForWhisper availlable for public channels.


merijndk:
Are there any plans on making .whisper and .listenForWhisper availlable for public channels. Would be really helpfull for chatrooms without user authentication. Or is there another way to send messages from client to other clients without a new broadcast event from the server?

driesvints:
We don't have plans ourselves but you're free to attempt a pr.

PR #297, opened 2021-01-17
[1.x] Support .whisper() on public channels

rennokki:
Related to #294


taylorotwell:
From Pusher docs: Client events can only be triggered on private and presence channels because they require authentication

PR #298, opened 2021-01-23
Add "interceptorUrls" option

aprivette:
The interceptorUrls option would allow users to apply the X-Socker-Id header only on requests to certain URLs.  This is an alternative to disabling it entirely using withoutInterceptors, which I’m currently doing in a project to prevent CORS errors from some third party APIs which don’t allow the header.

Omitting the interceptorUrls option falls back to the original behavior of adding the X-Socket-Id header to all requests.  The withoutInterceptors option will override this option.

Example usage:

```
new Echo({
    interceptorUrls: ['https://wants-x-socket-headers.com']
    ...
});
```

### DIFF
--- a/src/echo.ts
+++ b/src/echo.ts
@@ -111,9 +111,9 @@ export default class Echo {
     shouldApplySocketIdHeader(url: string): boolean {
         const origin = new URL(url).origin;
 
-        return this.socketId()
-            && (!('interceptorUrls' in this.options)
-            || this.options.interceptorUrls.includes(origin));
+        return (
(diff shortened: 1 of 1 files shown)


taylorotwell:
Just use a different axios instance for those URLs? This one is intended for making requests to your own application.

amiranda-fa:
Isn't it a valid use case to make requests to your own server and also have jQuery being used to make requests to external services where you'd like to disable the injection of the header?
You might even include a vendor's library that uses jQuery and makes requests and you don't have control over it.
What would be a good alternative for that use case?

PR #331, opened 2022-03-14
Bump packages versions

lucasmichot:
This PR bumps packages versions
This also gets rid of all vulnerabilities warning

---

It might be worth setting up [Dependabot](https://docs.github.com/en/code-security/supply-chain-security/managing-vulnerabilities-in-your-projects-dependencies/configuring-dependabot-security-updates) for this repository


driesvints:
Heya, thanks. @lucasmichot afaik Dependabot only works with .lock files, something this repo doesn't have.

Some bumps like "@babel/preset-env" aren't necessary because the current constrains already install these versions.

Did you try out Echo and all its features from the docs to make sure everything still works as expected?

PR #333, opened 2022-04-06
add Channel, PresenceChannel class to typescript definition exports

steve3d:
I've created a [issue](../issues/332 ) about the ` Channel`, `PresenceChannel` class should be also exported to typescript definition,

and here is the simple fix for that.


igorrynkovoy:
That is very good idea... (NO)
Now try to install this package using NPM (also check version) and use it reading Laravel DOC.
@steve3d 
Good luck!

PR #345, opened 2022-06-29
[1.x] Introduce root index for exporting

timacdonald:
This PR introduces an index file in the root of the package to deal with exporting. This is common convention for packages and is how we seem to do things across the Laravel ecosystem.

This also pairs with https://github.com/laravel/echo/pull/344 where we introduce a dedicated index for IIFE builds and removes what seems like arbitrary exports from the `echo.ts` file, just to support TS - where is this is a dedicated place to export things for TS etc.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -2,7 +2,7 @@ import babel from '@rollup/plugin-babel';
 import typescript from 'rollup-plugin-typescript2';
 
 export default {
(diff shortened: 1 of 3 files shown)


timacdonald:
Marked as a draft as it will conflict with the other PR. Will adjust once we merge the other one.

PR #340, opened 2022-06-02
User Authentication

rennokki:
## Context

- PR: https://github.com/laravel/framework/pull/42531
- `signin()` method: https://pusher.com/docs/channels/using_channels/user-authentication/#sign-in

## Caveats

This version requires Pusher JS SDK 7.1+ (https://github.com/pusher/pusher-js/releases/tag/v7.1.0-beta)

This PR needs to be merged before:
- Documentation: https://github.com/laravel/docs/pull/7965

## Breaking Changes

None.

PR #346, opened 2022-07-20
Bearer token support for User Authentication

erikn69:
## Context
* PR: [[9.x] User authentication for Pusher framework#42531](https://github.com/laravel/framework/pull/42531)
* `signin()` method: https://pusher.com/docs/channels/using_channels/user-authentication/#sign-in
* Complement of  https://github.com/laravel/echo/pull/340

## Breaking Changes
None.


jose123v:
it seems that only the title is enough to know what happens, maybe here they are too lazy to read two lines of code

PR #348, opened 2022-07-25
Adds Turbo Interceptor to add X-Socket-Id header on Turbo requests

tonysm:
### Added

- Adds a request interceptor to add the `X-Socket-Id` header to all Turbo requests by listening to the `turbo:before-fetch-request` event ([docs)](https://turbo.hotwired.dev/reference/events)

### DIFF
--- a/src/echo.ts
+++ b/src/echo.ts
@@ -121,6 +121,10 @@ export default class Echo {
         if (typeof jQuery === 'function') {
             this.registerjQueryAjaxSetup();
         }
+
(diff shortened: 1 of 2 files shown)


tonysm:
I've been using this in my [Turbo demo app](https://github.com/tonysm/turbo-demo-app/blob/main/resources/js/bootstrap.js#L22) for a while, thought it was be a good piece to submit to upstream. One caveat: the Turbo object must be imported and bound to the `window` object before Echo is initialized.

PR #356, opened 2022-09-15
DX improvments by adding typehints

nandi95:
This is a non-breaking change.
This will help users set up quicker by typehinting options.
There is no type testing set up nor there is much to test given no features have been added.
Please see comments below.

I would version this as a minor given the error added and the method on the null channel.

### DIFF
--- a/src/connector/connector.ts
+++ b/src/connector/connector.ts
@@ -5,16 +5,16 @@ type CustomConnector = new (options: Options) => Connector;
 type Broadcaster = 'socket.io' | 'pusher' | 'null' | CustomConnector
 
 export interface Options {
-    auth: {
+    auth?: {
         headers: Record<string, string>
     },
-    userAuthentication: {
-        endpoint: string,
-        headers: Record<string, string>,
+    userAuthentication?: {
+        endpoint?: string,
+        headers?: Record<string, string>,
     },
-    broadcaster: Broadcaster,
-    csrfToken: string | null,
-    bearerToken: string | null,
+    broadcaster?: Broadcaster,
+    csrfToken?: string | null,
+    bearerToken?: string | null,
     [K: string]: unknown
 }
 
--- a/src/connector/connector.ts
+++ b/src/connector/connector.ts
@@ -1,10 +1,28 @@
 import { Channel, PresenceChannel } from './../channel';
 
+type CustomConnector = new (options: Options) => Connector;
+
+type Broadcaster = 'socket.io' | 'pusher' | 'null' | CustomConnector
+
+export interface Options {
+    auth: {
+        headers: Record<string, string>
+    },
+    userAuthentication: {
+        endpoint: string,
+        headers: Record<string, string>,
+    },
+    broadcaster: Broadcaster,
+    csrfToken: string | null,
+    bearerToken: string | null,
+    [K: string]: unknown
+}
+
 export abstract class Connector {
     /**
      * Default connector options.
      */
-    private _defaultOptions: any = {
+    private _defaultOptions: Options = {
         auth: {
             headers: {},
         },
@@ -29,7 +47,7 @@ export abstract class Connector {
     /**
      * Create a new class instance.
      */
-    constructor(options: any) {
+    constructor(options: Options) {
         this.setOptions(options);
         this.connect();
     }
@@ -37,7 +55,7 @@ export abstract class Connector {
     /**
      * Merge the custom options with the defaults.
      */
-    protected setOptions(options: any): any {
+    protected setOptions(options: Options): any {
         this.options = Object.assign(this._defaultOptions, options);
 
         let token = this.csrfToken();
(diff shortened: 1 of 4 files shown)


src/connector/connector.ts
```diff
@@ -1,10 +1,28 @@
...
 
+type CustomConnector = new (options: Options) => Connector;
+
+type Broadcaster = 'socket.io' | 'pusher' | 'null' | CustomConnector
+
+export interface Options {
+    auth?: {
+        headers: Record<string, string>
+    },
+    userAuthentication?: {
```
nandi95, line 11: Haven't set up fields that are isn't actually referenced in this package. For service specific options the user can import typing from the service's package and use an intersection of types.

src/connector/connector.ts
```diff
@@ -83,6 +101,11 @@ export abstract class Connector {
      */
     abstract connect(): void;
 
+    /**
+     * Listen for an event on a channel instance.
+     */
+    abstract listen(name: string, event: string, callback: CallableFunction): Channel;
```
nandi95, line 107: This seems to have been present in all supplied connectors and this would be a basic requirement/naming for any different provider's connector.

src/connector/null-connector.ts
```diff
@@ -38,6 +38,13 @@ export class NullConnector extends Connector {
         return new NullPrivateChannel();
     }
 
+    /**
+     * Get a private encrypted channel instance by name.
+     */
+    encryptedPrivateChannel(name: string): NullPrivateChannel {
```
nandi95, line 44: Added so it's has better feature parity with the pusher connector.

nandi95: Not sure what the socket.io equivalent...

src/connector/socketio-connector.ts
```diff
@@ -50,7 +50,7 @@ export class SocketIoConnector extends Connector {
     /**
      * Listen for an event on a channel instance.
      */
-    listen(name: string, event: string, callback: Function): SocketIoChannel {
+    listen(name: string, event: string, callback: CallableFunction): SocketIoChannel {
```
nandi95, line 53: has better typing when it comes to `call`, `bind`, `apply`

nandi95: I can update the rest of the package to `CallableFunction` if desired.

src/echo.ts
```diff
@@ -1,24 +1,29 @@
...
+
+interface EchoOptions extends Options {
+    withoutInterceptors?: boolean;
+}
 
 /**
  * This class is the primary API for interacting with broadcasting.
  */
-export default class Echo {
+export default class Echo<O extends Record<PropertyKey, unknown>> {
```
nandi95, line 12: user can pass in the typings from providers such as [pusher](https://github.com/pusher/pusher-js/blob/7024cc8e1b68e5357b8ecacc9a0650b1033d6a07/src/core/options.ts#L13)

src/connector/connector.ts
nandi95: In fact `authEndpoint` for example is deprecated in pusher.

src/echo.ts
```diff
@@ -46,6 +51,8 @@ export default class Echo {
             this.connector = new NullConnector(this.options);
         } else if (typeof this.options.broadcaster == 'function') {
             this.connector = new this.options.broadcaster(this.options);
+        } else {
+            throw new Error('Broadcaster option not set when initialising Echo.');
```
nandi95, line 55: This is a bit nicer than the user calling a method on undefined as the connector has not been set.

src/echo.ts
```diff
@@ -95,7 +102,7 @@ export default class Echo {
      * Get a private encrypted channel instance by name.
      */
     encryptedPrivate(channel: string): Channel {
-        return this.connector.encryptedPrivateChannel(channel);
+        return (this.connector as NullConnector | PusherConnector).encryptedPrivateChannel(channel);
```
nandi95, line 105: Not sure why this is here socket.io connector is not implementing it. (The user can always just access `echo.connector`)

taylorotwell:
Thanks for your pull request to Laravel!

Unfortunately, I'm going to delay merging this code for now. To preserve our ability to adequately maintain the framework, we need to be very careful regarding the amount of code we include. 

If possible, please consider releasing your code as a package so that the community can still take advantage of your contributions! 

If you feel absolutely certain that this code corrects a bug in the framework, please "**@**" mention me in a follow-up comment with further explanation so that GitHub will send me a notification of your response.

nandi95:
@taylorotwell 
To respond, this kind of code if anything, it helps maintaining the codebase. In this case, I could release a `@types/laravel-echo` but it would be odd, given that files are already are typescript files and it would mean it's not updated along with echo itself.

nandi95:
@jessarcher Any chance this could be looked at again?

PR #372, opened 2023-02-07
Add Whisper typescript Prototype

### LINKED ISSUE #371: PresenceChannel interface missing 'Whisper' method
- Echo Version: 1.15.0
- Laravel Version: 9.19.0
- PHP Version: 8.17
- NPM Version: 8.12.2
- Node Version: 16.15.1

### Description:

The 'Whisper' method is not included in the type for PresenceChannel. Upon looking into the interface for PresenceChannel in laravel-echo, I can see that indeed, there isn't an interface method there.
### Steps To Reproduce:

In TypeScript, create an Echo instance 'Echo'
typeup:
Echo.join('room').whisper('test, () => {}).

You'll see an error as 'whisper' is not included in type 'PresenceChannel'

DellanX:
Closes #371, by adding the prototype to presence-channel.ts

### DIFF
--- a/src/channel/presence-channel.ts
+++ b/src/channel/presence-channel.ts
@@ -18,4 +18,9 @@ export interface PresenceChannel extends Channel {
      * Listen for someone leaving the channel.
      */
     leaving(callback: Function): PresenceChannel;
+    
+    /**
+     * Fire a whisper event to clients present in the channel.
+     */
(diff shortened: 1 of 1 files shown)


driesvints:
Seems this breaks the test suite. Please fix the tests and resend this. Thanks

PR #390, opened 2024-02-02
Refactor channels names(only readability fix)

erikn69:
This PR only adds better readability,
avoid repeating the same code many times
Is there any reason to repeat names like that?

```ts
//before
if (!this.channels['private-' + name]) {
    this.channels['private-' + name] = new PusherPrivateChannel(this.pusher, 'private-' + name, this.options);
}
//after
name = 'private-' + name;
if (!this.channels[name]) {
    this.channels[name] = new PusherPrivateChannel(this.pusher, name, this.options);
}
```


taylorotwell:
If there is no bug I think it is fine.

PR #396, opened 2024-04-08
Throw error on unsupported broadcaster

SuperDJ:
Added error messages for unsupported broadcaster


src/echo.ts
```diff
@@ -40,15 +40,22 @@ export default class Echo {
...
+            return;
         } else if (this.options.broadcaster == 'null') {
             this.connector = new NullConnector(this.options);
+            return;
         } else if (typeof this.options.broadcaster == 'function') {
             this.connector = new this.options.broadcaster(this.options);
+            return;
         }
+
+        throw new Error(`Broadcaster ${typeof this.options.broadcaster} ${this.options.broadcaster} is not supported`);
```
driesvints, line 58: Just put this in an `else`?

Issue #408, opened 2025-01-14
Latest Update broke auth


robgordon89:
### Echo Version

1.18.0

### Laravel Version

11

### PHP Version

8.3

### NPM Version

8.19.4

### Description

The latest release 1.18.0 broke channel auth please see attached, reverting back to 1.17.1 resolved it, I am not sure what the issue is i didnt have time to dig into it 👍 

![Image](https://github.com/user-attachments/assets/fba3b7ca-9890-4686-82ec-a3472688edea)

Thanks
Bob

### Steps To Reproduce

Install 1.18.0

crynobone:
Hey there, thanks for reporting this issue.

We'll need more info and/or code to debug this further. Can you please create a repository with the command below, commit the code that reproduces the issue *as one separate commit* on the main/master branch and share the repository here?

Please make sure that you have the [latest version of the Laravel installer](https://github.com/laravel/installer) in order to run this command. Please also make sure you have both Git & [the GitHub CLI tool](https://cli.github.com/) properly set up.

    laravel new bug-report --github="--public"

Do not amend and create a separate commit with your custom changes. After you've posted the repository, we'll try to reproduce the issue.

Thanks!

mandatoryGithubUser:
How can I use a custom authorizer that would post with axios? I'm using Breeze Nextjs with Breeze API only (Laravel), Sanctum requires CSRF tokens and that'd be saved when the auth happens. 

In `echo: 1.17.1`. I could do this

```
/// Axios (Default from Breeze Next)
```
import Axios from 'axios'

const axios = Axios.create({
    baseURL: process.env.NEXT_PUBLIC_BACKEND_URL,
    headers: {
        'X-Requested-With': 'XMLHttpRequest',
    },
    withCredentials: true,
    withXSRFToken: true
})

export default axios
```

/// echo.ts 
        return new Echo({
            ...,
            authorizer: channel => {
                return {
                    authorize: (socketId, callback) => {
                        axios
                            .post(`/broadcasting/auth`, {
                                socket_id: socketId,
                                channel_name: channel.name,
                            })
                            .then(response => {
                                callback(null, response.data)
                            })
                            .catch(error => {
                                callback(error)
                            })
                    },
                }
            },
      })
```

But now, my broadcasting/auth url defaults to the frontend url, instead of the backend url I set. If I set channelAuthorization endpoint, then there's no longer a CSRF token that my axios had when logging in.

filipescaglia:
[...]

PR #409, opened 2025-01-20
fix(connector): authEndpoint backward compatibility

021-projects:
Fix https://github.com/laravel/echo/issues/408

Issue source: https://github.com/pusher/pusher-js/blob/37b057c45c403af27c79303123344c42f6da6a25/src/core/config.ts#L151


021-projects:
Bad implementation, will open a new PR

PR #410, opened 2025-01-20
fix(connector): authEndpoint backward compatibility

021-projects:
Fix https://github.com/laravel/echo/issues/408

Issue source: https://github.com/pusher/pusher-js/blob/37b057c45c403af27c79303123344c42f6da6a25/src/core/config.ts#L151

### DIFF
--- a/src/connector/connector.ts
+++ b/src/connector/connector.ts
@@ -5,6 +5,10 @@ export abstract class Connector<TPublic extends Channel, TPrivate extends Channe
      * Default connector options.
      */
     private _defaultOptions: any = {
+        /**
+         * @deprecated Use `channelAuthorization.endpoint` instead.
+         */
+        authEndpoint: '/broadcasting/auth',
         channelAuthorization: {
             endpoint: '/broadcasting/auth',
             headers: {},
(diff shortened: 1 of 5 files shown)


kirkbushell:
As good as it is to see this PR, this appears to break the standards and conventions set up the package, using ES6 classes instead of functions to manage the complexity. It would appear that moving those options out of that class is a mistake, and should be managed as part of the Connector API, as that manages the options.

I think it's a relatively straight-forward change - if authorizer is present as part of the options, use that, else default to the other options for connectivity.

021-projects:
> As good as it is to see this PR, this appears to break the standards and conventions set up within package, using ES6 classes instead of functions to manage the complexity. It would appear that moving those options out of that class is a mistake, and should be managed as part of the Connector API, as that manages the options.

In essence, options are still managed by classes - they just use utilities to improve code organization, which already happened with the `isConstructor` utility. Do you think this is a fundamental change? It seems like it's just code reuse. This makes the `PusherConnector` look much cleaner while still maintaining the same functionality. Also, the authorization options were moved to `PusherConnector` because `Connector` is an abstraction that is also used by other connectors that do not support Pusher options, which means that the abstraction only has code for one specific case. I thought this needed to be improved.


> I think it's a relatively straight-forward change - if authorizer is present as part of the options, use that, else default to the other options for connectivity.

It's not just the `authorizer`, but several other options that stop working when the `channelAuthorization` property is provided, so the commit https://github.com/laravel/echo/commit/fad559947c1796b502b4842a8cc8de0b59884741 essentially breaks the authorization functionality entirely, despite the fact that the properties have been deprecated for almost three years.

taylorotwell:
This feels like a really big change?

taylorotwell:
@ProjektGopher can you take a look at this if you have a chance?

021-projects:
> This feels like a really big change?

Not really big, just some refactoring to move Pusher options to the Pusher connector and fix backward compatibility

taylorotwell:
Reverted the original PR for now. cc @ProjektGopher

PR #439, opened 2025-05-21
[Feature] Add Sockudo to Broadcasting driver

countradooku:
This PR introduces add sockudo support to laravel echo

### DIFF
--- a/packages/laravel-echo/src/echo.ts
+++ b/packages/laravel-echo/src/echo.ts
@@ -73,6 +73,11 @@ export default class Echo<T extends keyof Broadcaster> {
                 cluster: "",
                 broadcaster: "pusher",
             });
+        } else if (this.options.broadcaster === "sockudo") {
+            this.connector = new PusherConnector<"sockudo">({
+                ...this.options,
+                cluster: "",
+            });
         } else if (this.options.broadcaster === "socket.io") {
             this.connector = new SocketIoConnector(this.options);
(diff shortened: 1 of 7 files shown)


countradooku:
i close this as it works with pusher broadcaster

dam1r89:
Would it make sense to reopen this?

Recently Sockudo released a version 2 of their protocol and I believe it is necessary to use their client to benefit from the features.

countradooku:
well they closed my PR to the docs too..

countradooku:
i think laravle is made to work only with pusher compatible servers. it could need some more changes to work with sokcudo full features

dam1r89:
I was hoping that the Connection Recovery feature is something that works between websocket server and client without back-end laravel changes required.

countradooku:
@dam1r89 i will investigate this

dam1r89:
After replacing the old setup laravel-echo `pusher` broadcaster and **soketi** websocket server with **sockudo-js** and **sockudo** websocket server (and configuring it properly by: enabling connection recovery, setting buffer size an ttl) it worked without changing the pusher server `pusher/pusher-php-server`

```javascript
import Sockudo from "@sockudo/client";

window.Pusher = Sockudo; // ugly

const echoInstance = new Echo({
    broadcaster: 'pusher',
    protocolVersion: 2,
    connectionRecovery: true,
    /** rest of options */
```

Thanks @countradooku

PR #448, opened 2025-07-18
feat(vue): allow checking if echo is configured

hendrikheil:
This PR adds a simple additional helper function to the Vue integration of Laravel Echo. This is a useful addition for cases where you might want echo to be an optional addition to an app. This allows you to check if the application configured echo, so you can listen to events elsewhere in the application, without having to handle the exception that `useEcho` or `echo` will throw when unconfigured.

Example:
```
if (import.meta.client && isEchoConfigured()) {
  useEcho(...)
}
```

Prior to this change you have to either handle the configuration state check yourself or catch the error.


joetannenbaum (approved):
Adjusted the name of the function and added it to the React package as well 👍

PR #469, opened 2025-12-22
Add a connection status hook (discussion required)

nexxai:
## Problem

I want to be able to determine the current status of my connection to Reverb.  As it stands there isn't currently a way to reliably retrieve this status.

## Proposed Solution

I have added a `useConnectionStatus` hook that returns the current status, fetching it from the existing Pusher and Socket.IO libraries behind the scenes.

## Notes

I am not currently sold on the return values as strings, but I couldn't come up with anything better that clearly represented the various states.  I am *very open* to alternative suggestions; I just wanted to get something out there so that we could start the discussion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,16 @@
 # Release Notes
 
-## [Unreleased](https://github.com/laravel/echo/compare/v2.2.6...2.x)
+## [Unreleased](https://github.com/laravel/echo/compare/v2.2.7...2.x)
+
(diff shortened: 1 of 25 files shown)


joetannenbaum:
This looks super useful, thank you! I'll give this a proper review after the new year and let's do it up 👍

joetannenbaum:
This looks great, thank you! I ended up removing `getConnectionStatus`, users can just use `echo().connectionStatus()`. I also updated the `useConnectionStatus` hook to subscribe to status changes so that the hook reports changes in real time.

PR #472, opened 2026-01-08
fix(hooks): resolve performance issues and stale closure bugs in useEcho

pataar:
First of all, happy to receive any feedback. Also don't hesitate to ask for any clarifications.

### Problem

The `useEcho` hooks had several issues that could cause bugs in production:

1. **Stale closures** - callbacks captured outdated values because dependency arrays didn't include all referenced variables
2. **Unstable references** - `channel`, `events`, and the return object were recreated every render, causing unnecessary re-subscriptions and breaking downstream `useMemo`/`useCallback` in consumer components
3. **Array reference issues** - passing `['event1', 'event2']` as a prop would bust memoization even when contents were identical

### Changes

**`useEcho`**
- Memoize `channel` object based on `channelName` and `visibility`
- Memoize `events` using a derived string key to handle array reference instability
- Fix all `useCallback` dependency arrays to include actual dependencies (`events`, `callbackFunc`, etc.)
- Memoize return object so consumers get stable references
- Rename refs for clarity (`listening` → `listeningRef`, etc.)

**`useEchoNotification`**
- Add `memoizedCallback` wrapper to stabilize the notification handler
- Fix `cb` dependencies
- Fix effect to use `[listen, stopListening]` instead of manually concatenating arrays

**Types**
- Replace `any[]` with `DependencyList` from React

**ESLint**
- Add `eslint-plugin-react-hooks` and `@eslint-react/eslint-plugin`
- These catch the exact class of bugs this PR fixes

### Before/After
```tsx
// Before: this would cause stale closure bugs
useEcho('channel', ['event'], (data) => {
  console.log(someValue); // could be stale
}, [someValue]);

// After: properly updates when someValue changes
```

### Breaking Changes

None. API is unchanged.


packages/react/eslint.config.mjs
```diff
@@ -19,22 +23,6 @@ const config = [
         rules: {
             ...tsPlugin.configs.recommended.rules,
             ...tsPlugin.configs["recommended-requiring-type-checking"].rules,
-            "@typescript-eslint/ban-types": "off",
```
pataar, line 22: Removed these because the fixes in this PR make them unnecessary. Happy to revert if we'd rather keep them for consistency.

joetannenbaum:
Thank you!

pataar:
Thanks for merging!

PR #481, opened 2026-03-05
fix: useEchoNotification callback not firing after re-render

### LINKED ISSUE #467: useEchoNotification does not trigger my callback
### Echo Version

2.2.6

### Laravel Version

12.0

### PHP Version

8.4.15

### NPM Version

11.7.0

### Database Driver & Version

sqlite

### Description

I have two hooks listening to the same notification channel and logging a message when a notification is received.

```ts
const channel = `App.Models.User.${auth.user.id}`;

useEchoNotification(channel, () => console.log('user notification channel using useEchoNotification'));

useEcho(
    channel,
    '.Illuminate\\Notifications\\Events\\BroadcastNotificationCreated',
    () => console.log("user notification channel using useEcho"),
);
```

I expect to see two logs in the console, but only the second one (useEcho) is logged.

### Steps To Reproduce

A minimal reproduction is available here: https://github.com/DhiaBen20/laravel-echo-bug-report

Steps:
1. Clone the repository
2. Install dependencies
3. cp .env.example .env
4. php artisan key:gen
5. php artisan migrate
6. php artisan reverb:install
7. composer run dev
8. php artisan reverb:start
9. login and go to dashboard page and open the console.
10. inside tinker notify the currently authenticated user `User::first()->notify(new \App\Notifications\Test)` and you will see one log printed on the console instead of two

pataar:
Fixes #467

## The problem

`useEchoNotification` stops working after a re-render because the notification listener is never re-registered.

Both the React and Vue hooks had an `initialized` flag that was meant to prevent double-registration, but it also prevented *re*-registration after cleanup. Here's what happens:

```ts
// 1. First mount — works fine
listen()
  → listening = false, initialized = false
  → calls channel.notification(cb) ✅
  → sets listening = true, initialized = true

// 2. Callback changes (re-render) → effect cleanup runs
stopListening()
  → calls channel.stopListeningForNotification(cb)
  → sets listening = false
  // but initialized is still true!

// 3. Effect re-runs
listen()
  → listening = false ✓ (passes the guard)
  → initialized = true ✗ (skips notification registration!)
  → listener is gone forever
```

This means any component re-render that changes the callback reference will silently kill the notification listener. It also affects the manual `stopListening()` / `listen()` cycle shown in #467.

## The fix

Remove the `initialized` guard entirely. The `listening` flag already prevents double-registration — if `listening` is `true`, `listen()` is a no-op. After `stopListening()` sets it back to `false`, `listen()` correctly re-registers the notification callback.

This matches how the base `useEcho` hook already works — it only uses a `listening` guard, no separate `initialized` flag.

## Test plan

- [x] Updated existing "can manually start and stop listening" tests in both React and Vue to assert that `notification()` is called again after a stop/listen cycle
- [x] All existing tests pass (`pnpm test`)

### DIFF
--- a/packages/react/src/hooks/use-echo.ts
+++ b/packages/react/src/hooks/use-echo.ts
@@ -281,7 +281,6 @@ export const useEchoNotification = <
     }, [eventKey]);
 
     const listening = useRef(false);
-    const initializedRef = useRef(false);
 
     // eslint-disable-next-line react-hooks/exhaustive-deps
     const memoizedCallback = useCallback(callback, dependencies);
@@ -304,12 +303,9 @@ export const useEchoNotification = <
             return;
         }
 
-        if (!initializedRef.current) {
-            result.channel().notification(cb);
-        }
+        result.channel().notification(cb);
 
         listening.current = true;
-        initializedRef.current = true;
     }, [cb, result]);
 
     const stopListening = useCallback(() => {
--- a/packages/react/tests/use-echo.test.ts
+++ b/packages/react/tests/use-echo.test.ts
@@ -1222,8 +1222,8 @@ describe("useEchoNotification hook", async () => {
 
         result.current.listen();
 
-        // notification should still only be called once due to initialized check
-        expect(channel.notification).toHaveBeenCalledTimes(1);
+        // notification should be re-registered after stop + listen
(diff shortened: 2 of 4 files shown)


joetannenbaum:
Thank you!

PR #488, opened 2026-04-15
Refactor useConnectionStatus

ezra-obiwale:
Even though `const status = ref<ConnectionStatus>(echo().connectionStatus())` exists, the connection is not correctly set on mounted.

This PR sets the status on mounted as well. Used a single echo instance while it.


joetannenbaum:
Thank you!

PR #490, opened 2026-04-16
Support listener context on Pusher subscriptions

jlswanson28694:
Pusher has the ability to add context as the third argument when binding and unbinding callbacks, which conveniently lets me remove multiple bound callbacks at once using `channel.unbind(undefined, undefined, context)`. This PR adds support for context when using Pusher channels, and adds a method `stopListeningForContext` to remove all listeners for the given context.

I've also added a `PusherChannel` test for this feature. In order to properly test `PusherChannel`, the test starts a bare-bones WebSocketServer with the `ws` library. I'm no expert on frontend testing, so feel free to clean it up as you see fit.

### DIFF
--- a/packages/laravel-echo/package.json
+++ b/packages/laravel-echo/package.json
@@ -40,6 +40,7 @@
         "@babel/preset-env": "^7.26.7",
         "@types/jquery": "^3.5.32",
         "@types/node": "^20.0.0",
+        "@types/ws": "^8.18.1",
         "@typescript-eslint/eslint-plugin": "^8.21.0",
         "@typescript-eslint/parser": "^8.21.0",
         "axios": "^1.15.0",
@@ -51,7 +52,8 @@
         "typescript": "^5.7.0",
         "vite": "^6.4.2",
         "vite-plugin-dts": "^4.5.3",
-        "vitest": "^3.1.2"
+        "vitest": "^3.1.2",
+        "ws": "^8.20.0"
     },
     "peerDependencies": {
         "pusher-js": "*",
--- a/packages/laravel-echo/src/channel/pusher-channel.ts
+++ b/packages/laravel-echo/src/channel/pusher-channel.ts
@@ -66,8 +66,8 @@ export class PusherChannel<
     /**
      * Listen for an event on the channel instance.
      */
-    listen(event: string, callback: CallableFunction): this {
-        this.on(this.eventFormatter.format(event), callback);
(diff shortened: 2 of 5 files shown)


joetannenbaum:
Thanks for the PR and the detailed write-up. I went back and forth on this one, but I think we're going to pass for now. A couple of reasons:

- Most cleanup scenarios in Echo apps are already covered by `Echo.leaveChannel(name)`, and for the rarer case where you need finer-grained cleanup on a shared channel, `channel.subscription.bind/unbind` is exposed directly and supports context natively.
- We try to keep the cross-driver API surface consistent, and a Pusher-only `stopListeningForContext` becomes a long-term maintenance signal ("why doesn't socket.io support this?") that's hard to take back once shipped.

If you find this comes up a lot in practice and you can show real-world usage that `leaveChannel` doesn't cover well, I'm open to revisiting, but as a net-new public API I don't think the win is there yet.

jlswanson28694:
@joetannenbaum Let me add more context by attempting to explain my use-case.

I have a Vue component `ReportDownload` that renders a button that, when clicked, will tell the backend and add a new job to the queue to generate a large report. The Vue component then connects to Reverb on the `reports.{userId}` channel, displays a "processing" toast, and waits for a `ReportReady` event (or `ReportFailed`).

I'm allowing the user to generate multiple reports like this simultaneously. As a result, if one report completes, I can't run `leaveChannel` without killing the connection for other in-progress reports. This is a simplified version of what my Vue component looks like using this new API:
```ts
new Promise<ReportReadyData>((resolve, reject) => {
    this.channel = this.echo.private(BroadcastChannels.reports(user.value.id))
        .listen(
            events.ReportReady,
            (e: ReportReadyData) => {
                if (e.report_id === this.report_id) resolve(e);
            },
            this.report_id
        )
        .listen(
            events.ReportFailed,
            (e: ReportIdentifier) => {
                if (e.report_id === this.report_id) reject();
            }, 
            this.report_id
        )
        .error(reject, this.report_id);
})
.finally(() => {
    this.channel.stopListeningForContext(this.report_id);
});
```

Thought it looked pretty clean. As you said, `channel.subscription.bind/unbind` does work, but that was definitely more verbose. The `error` callback in particular translates to `channel.subscription.bind("pusher:subscription_error", reject, this.report_id)`. 

Looking at this again, I suppose I could open a new channel for each report instead, but this is just to show you where my head was when I wrote the PR. No worries if you don't think it's worth adding.

Issue #474, opened 2026-01-13
No official way to access socketId when using Reverb + toOthers() with @laravel/echo-react


tranvantruongktech:
### Echo Version

2.2.7

### Laravel Version

12

### PHP Version

8.4

### NPM Version

11.4.2

### Description

I am using Laravel 12, Laravel Reverb, and @laravel/echo-react to build a real-time chat feature.

Broadcasting itself works correctly, however I am running into a limitation when using broadcast()->toOthers().

Laravel requires the client to send the X-Socket-Id header so the broadcaster can exclude the sender socket.
With classic Echo usage (window.Echo), this is straightforward via Echo.socketId().

However, when using @laravel/echo-react:

configureEcho() does not expose Echo globally

window.Echo is always undefined

There is no documented or supported way to retrieve socketId outside of React hooks

There is no helper like getEcho() or similar to access the Echo instance

As a result, users must rely on workarounds (e.g. storing socketId in a global variable from a component) to make toOthers() function correctly.

### Steps To Reproduce

**Expected behavior**

One of the following (or similar) would make integration much easier:

An official API to access the Echo instance or socketId

e.g. getEcho()

or getSocketId()

Optional exposure of the Echo instance (configurable)

Official documentation describing how toOthers() should be handled when using @laravel/echo-react with Reverb

**Actual behavior**

window.Echo is undefined

No supported way to access socketId

broadcast()->toOthers() cannot work correctly without custom hacks

**Environment**

Laravel: 12.x

Broadcasting driver: Reverb

Frontend: React + Inertia

Package: @laravel/echo-react

Realtime use case: chat messaging

caiokawasaki:
Just hit the same problem right now

joetannenbaum:
We expose an `echo` helper to access the global instance:

```ts
import { echo } from '@laravel/echo-react';

echo().socketId()
```

HarbourDesign:
They need to update the docs with this. Here's an example with Vue .ts files and a Wayfinder route that works directly without a Global setup:  
```   
sendForm.submit(store(), {
        headers: {
            'X-Socket-ID': echo().socketId() ?? '',
        },
        onSuccess: () => {
            console.log('Message sent successfully');
        },
        onError: (errors) => {
            console.error('Failed to send message:', errors);
        },
    });
```

PR #498, opened 2026-05-14
feat: add useSocketId() hook to React, Vue and Svelte adapters

hectorgrecco-onerail:
## Summary

Adds a `useSocketId()` reactive hook/composable/rune to all three framework adapters, closing the gap described in #474.

The existing `echo()` function is already exported from all adapters and can be used directly for interceptor setup:

```ts
import { echo } from '@laravel/echo-react';

axios.interceptors.request.use((config) => {
    const socketId = echo().socketId();
    if (socketId) config.headers['X-Socket-Id'] = socketId;
    return config;
});
```

However, there was no reactive primitive for component usage that automatically updates when the connection reconnects and receives a new socket ID. This PR fills that gap.

### React
```ts
import { useSocketId } from '@laravel/echo-react';

const socketId = useSocketId(); // string | undefined, re-renders on reconnect
```

### Vue
```ts
import { useSocketId } from '@laravel/echo-vue';

const socketId = useSocketId(); // ComputedRef<string | undefined>
```

### Svelte
```ts
import { useSocketId } from '@laravel/echo-svelte';

const socketId = useSocketId(); // () => string | undefined (getter, Svelte 5 runes pattern)
```

## Implementation

`useSocketId()` builds on the existing `useConnectionStatus()` in each adapter — it subscribes to `onConnectionChange` for free, so when the connection cycles (and a new socket ID is assigned), the value updates automatically. No new subscriptions or state management.

## Test plan

- [ ] `packages/react`: 61 tests pass, 1 pre-existing skip
- [ ] `packages/vue`: 45 tests pass, 2 pre-existing skips
- [ ] `packages/svelte`: 32 tests pass

Related: #474

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/packages/react/src/hooks/use-echo.ts
+++ b/packages/react/src/hooks/use-echo.ts
@@ -416,3 +416,15 @@ export const useConnectionStatus = (): ConnectionStatus => {
 
     return status;
 };
+
+/**
+ * Hook to get the current WebSocket socket ID.
+ * Returns undefined until the connection handshake completes.
+ * Updates automatically when the connection reconnects.
+ *
+ * @returns string | undefined - The current socket ID, or undefined if not yet connected
+ */
+export const useSocketId = (): string | undefined => {
+    useConnectionStatus();
+    return echo().socketId();
+};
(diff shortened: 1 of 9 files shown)


hectorgrecco-onerail:
Closing — opening from correct account.

PR #499, opened 2026-05-14
feat: add useSocketId() hook to React, Vue and Svelte adapters

### LINKED ISSUE #474: No official way to access socketId when using Reverb + toOthers() with @laravel/echo-react

hectorgrecco:
## Summary

Adds a `useSocketId()` reactive hook/composable/rune to all three framework adapters, closing the gap described in #474.

The existing `echo()` function is already exported from all adapters and can be used directly for interceptor setup:

```ts
import { echo } from '@laravel/echo-react';

axios.interceptors.request.use((config) => {
    const socketId = echo().socketId();
    if (socketId) config.headers['X-Socket-Id'] = socketId;
    return config;
});
```

However, there was no reactive primitive for component usage that automatically updates when the connection reconnects and receives a new socket ID. This PR fills that gap.

### React
```ts
import { useSocketId } from '@laravel/echo-react';

const socketId = useSocketId(); // string | undefined, re-renders on reconnect
```

### Vue
```ts
import { useSocketId } from '@laravel/echo-vue';

const socketId = useSocketId(); // ComputedRef<string | undefined>
```

### Svelte
```ts
import { useSocketId } from '@laravel/echo-svelte';

const socketId = useSocketId(); // () => string | undefined (getter, Svelte 5 runes pattern)
```

## Implementation

`useSocketId()` builds on the existing `useConnectionStatus()` in each adapter — it subscribes to `onConnectionChange` for free, so when the connection cycles (and a new socket ID is assigned), the value updates automatically. No new subscriptions or state management.

## Test plan

- [ ] `packages/react`: 61 tests pass, 1 pre-existing skip
- [ ] `packages/vue`: 45 tests pass, 2 pre-existing skips
- [ ] `packages/svelte`: 32 tests pass

Closes #474

### DIFF
--- a/packages/react/src/hooks/use-echo.ts
+++ b/packages/react/src/hooks/use-echo.ts
@@ -417,13 +417,6 @@ export const useConnectionStatus = (): ConnectionStatus => {
     return status;
 };
 
-/**
- * Hook to get the current WebSocket socket ID.
- * Returns undefined until the connection handshake completes.
- * Updates automatically when the connection reconnects.
- *
- * @returns string | undefined - The current socket ID, or undefined if not yet connected
- */
 export const useSocketId = (): string | undefined => {
     useConnectionStatus();
     return echo().socketId();
--- a/packages/react/src/hooks/use-echo.ts
+++ b/packages/react/src/hooks/use-echo.ts
@@ -416,3 +416,15 @@ export const useConnectionStatus = (): ConnectionStatus => {
 
     return status;
 };
+
+/**
+ * Hook to get the current WebSocket socket ID.
+ * Returns undefined until the connection handshake completes.
+ * Updates automatically when the connection reconnects.
+ *
+ * @returns string | undefined - The current socket ID, or undefined if not yet connected
+ */
+export const useSocketId = (): string | undefined => {
+    useConnectionStatus();
+    return echo().socketId();
+};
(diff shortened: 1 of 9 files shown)


joetannenbaum:
Thanks for this! Did a light refactor but otherwise looks good.

Issue #482, opened 2026-03-11
useEchoNotification callback not invoked in React StrictMode


giuseppe-osenda:
### Echo Version

2.3.0

### Laravel Version

12.53.0

### PHP Version

8.5

### NPM Version

11.10.0

### Description

I couldn't explain why useEchoNotification callback didn't get called so i asked AI and this is what has found after digging through the file **node_modules/@laravel/echo-react/dist/index.js**:

**### AI Output**
useEchoNotification silently fails to bind its notification listener when the app is wrapped in <React.StrictMode>. The channel subscription succeeds and events arrive (visible in the network tab), but the callback is never invoked.

useEcho works correctly under the same conditions.

**Root cause**

In useEchoNotification, the listen function (h) uses a hasSubscribed ref (g) to guard against duplicate .notification() calls:

`const h = u(() => {
    c.current || (g.current || r.channel().notification(b), c.current = !0, g.current = !0);
}, [b, r]);`

The cleanup function resets c (isListening) but not g (hasSubscribed):
`const a = u(() => {
    c.current && (r.channel().stopListeningForNotification(b), c.current = !1);
}, [b, r]);`

In StrictMode, React runs effects, cleans them up, and re-runs them. During cleanup, stopListeningForNotification(b) unbinds the handler and c is reset to false. Meanwhile, useEcho (internally) leaves and re-creates the channel. On the second mount, h() is called again but g.current is still true (refs persist across StrictMode remounts), so r.channel().notification(b) is skipped. The new channel has no notification listener bound.

By contrast, useEcho handles this correctly because its useEffect unconditionally re-binds listeners via o() on every mount cycle.

**### END AI OUTPUT**

The workaround is to use useEcho directly with the notification event name:

`useEcho(
    channelName,
    '.Illuminate\\Notifications\\Events\\BroadcastNotificationCreated',
    (notification) => {
        if (notification.type === 'your_notification_type') {
            // handle notification
        }
    },
);`

### Steps To Reproduce

1. Wrap your app in <React.StrictMode> (default for create-react-app / Vite React templates)
2. Use useEchoNotification to listen for notifications on a private channel
3. Broadcast a notification to that channel
4. Observe: channel subscription succeeds, event arrives in the WebSocket, but the callback is never called

naveentehrpariya:
I looked into this and can confirm the root cause matches the report, but it is already fixed on current releases:

[...]

PR #522, opened 2026-06-12
Add StrictMode regression tests for useEchoNotification (#482)

naveentehrpariya:
Related to #482.

While investigating #482 (`useEchoNotification` callback never invoked under `React.StrictMode`), I found the bug no longer reproduces on `2.x`:

- **v2.3.0** (the reported version): on StrictMode's mount → cleanup → remount cycle, the cleanup unbound the notification listener via `stopListeningForNotification()`, but the `initializedRef` guard in `listen()` prevented the remount from ever re-binding it — the channel ended up with no notification listener, matching the report.
- **`2.x` / v2.3.5+**: the `initializedRef` guard was removed in the channel-subscription rework, so the remount re-binds correctly and the callback fires.

Since nothing covered this, this PR adds regression tests. The mock implements real bind/unbind semantics (a `Set` of bound listeners mutated by `notification()` / `stopListeningForNotification()`), so the assertion checks what is actually still bound after the StrictMode cycle rather than just that `notification()` was called at some point:

- `useEchoNotification` still has a bound listener and invokes the callback after the StrictMode remount — **fails against the v2.3.0 implementation, passes on `2.x`** (verified by swapping in the old `use-echo.ts`).
- A `useEcho` control test covering the same cycle for regular events.

#482 itself can likely be closed — the fix has shipped, the reporter was on 2.3.0.

### DIFF
--- a/packages/react/tests/use-echo-strict-mode.test.ts
+++ b/packages/react/tests/use-echo-strict-mode.test.ts
@@ -1,0 +1,116 @@
+import { act, renderHook } from "@testing-library/react";
+import { StrictMode } from "react";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const getEchoModule = async () => import("../src/hooks/use-echo");
+const getConfigModule = async () => import("../src/config/index");
+
+// Regression tests for https://github.com/laravel/echo/issues/482: in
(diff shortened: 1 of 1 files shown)


taylorotwell:
This feels like a lot of code. 😕

PR #530, opened 2026-07-01
Fix presence and whisper examples in echo-react Boost skill

### LINKED ISSUE #521: [echo-react] Boost skill examples: presence listeners registered outside useEffect; undefined user in whisper example
## Summary

The Boost skill shipped with `@laravel/echo-react` (`resources/boost/skills/echo-react-development/SKILL.blade.php`) contains two example bugs that coding agents copy verbatim into applications.

### 1. Presence-channel listeners registered outside `useEffect`

In the `useEchoPresence` example (lines ~112–114 of the published `SKILL.blade.php`):

```tsx
function ChatRoom({ roomId }: { roomId: number }) {
    const { channel } = useEchoPresence(`chat.${roomId}`, "NewMessage", (e) => {
        console.log(e.message);
    });

    channel().here((users) => console.log('Current users:', users));
    channel().joining((user) => console.log(`${user.name} joined`));
    channel().leaving((user) => console.log(`${user.name} left`));

    return <div>Chat room {roomId}</div>;
}
```

`channel().here()/.joining()/.leaving()` run in the component body, so every re-render registers duplicate listeners on the same channel. They also violate React's render-purity rules (flagged by `eslint-plugin-react-hooks` v7 and the React Compiler). Suggested fix:

```tsx
useEffect(() => {
    channel().here((users) => console.log('Current users:', users));
    channel().joining((user) => console.log(`${user.name} joined`));
    channel().leaving((user) => console.log(`${user.name} left`));
}, [channel]);
```

### 2. Whisper example references undefined `user` and whispers during render

In the client-events example (line ~159):

```tsx
function ChatInput({ roomId }: { roomId: number }) {
    const { channel } = useEcho(`chat.${roomId}`, ['update'], (e) => { ... });

    // Send typing indicator
    channel().whisper('typing', { name: user.name });
    ...
}
```

`user` is never declared (compile error as written), and the `whisper` call executes on every render rather than on a user action. Suggested fix: accept `user` as a prop and move the whisper into an event handler (e.g. an input's `onChange`), with `listenForWhisper` registered inside `useEffect`.

## Context

Found while syncing Boost skills into a Laravel 13 + React 19 app (`php artisan boost:update`). The framework-agnostic `laravel-echo` skill is unaffected — these examples exist only in the React skill. We're carrying a project-level `.ai/skills/` override in the meantime.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

lazerg:
The echo-react Boost skill had two example bugs that coding agents copy verbatim: the presence-channel `here`/`joining`/`leaving` listeners were registered in the component body instead of inside a `useEffect`, and the whisper example referenced an undefined `user` and whispered on every render. This wraps the presence listeners in a `useEffect`, passes `user` in as a prop, moves the `whisper` into an `onChange` handler, and registers `listenForWhisper` inside a `useEffect`.

Fixes #521

### DIFF
--- a/packages/react/resources/boost/skills/echo-react-development/SKILL.blade.php
+++ b/packages/react/resources/boost/skills/echo-react-development/SKILL.blade.php
@@ -102,6 +102,7 @@ function PostFeed() {
 ### Presence Channels
 
 @boostsnippet("Presence Channel Hook", "tsx")
+import { useEffect } from "react";
 import { useEchoPresence } from "@laravel/echo-react";
 
 function ChatRoom({ roomId }: { roomId: number }) {
@@ -109,9 +110,11 @@ function ChatRoom({ roomId }: { roomId: number }) {
         console.log(e.message);
     });
 
-    channel().here((users) => console.log('Current users:', users));
-    channel().joining((user) => console.log(`${user.name} joined`));
-    channel().leaving((user) => console.log(`${user.name} left`));
+    useEffect(() => {
+        channel().here((users) => console.log('Current users:', users));
+        channel().joining((user) => console.log(`${user.name} joined`));
+        channel().leaving((user) => console.log(`${user.name} left`));
+    }, [channel]);
 
     return <div>Chat room {roomId}</div>;
 }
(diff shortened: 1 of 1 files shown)


lazerg:
The `tests` and `format` failures are a pre-existing lockfile mismatch on `2.x`: `pnpm install --frozen-lockfile` errors on `jsdom` (`^29.1.1` in `pnpm-lock.yaml` vs `>=29.0.0` in `packages/react/package.json`) before anything runs, so it fails the same way on the base branch. This PR only edits the react skill doc.

PR #541, opened 2026-08-21
build: upgrade eslint to 10 and testing-library packages

pataar:
Follow-up to #540 — stacks on that branch, so review #540 first. The diff here shows only this commit once #540 lands.

**Upgrades**

| Package | From | To |
|---|---|---|
| `eslint` | 9.25.1 | 10.8.1 |
| `@typescript-eslint/{eslint-plugin,parser}` | 8.31.1 | 8.67.0 |
| `@eslint-react/eslint-plugin` | 2.5.1 | 5.18.6 |
| `eslint-plugin-react-hooks` | 7.0.1 | 7.1.1 |
| `@testing-library/svelte` | 5.3.1 | 5.4.2 |
| `@vue/test-utils` | 2.4.6 | 2.4.11 |
| `@testing-library/dom` | 10.4.0 | 10.4.1 |

The existing flat configs needed no migration. Every plugin peer already accepted `eslint ^10`.

**Three fixes the stricter rules require**

- `@typescript-eslint/no-unnecessary-type-assertion` flagged two casts that are now provably redundant: `socketio-connector` (`io()` already declares `opts?: Record<string, unknown>`) and the svelte `useEcho` rune (assignment target is already `Echo<BroadcastDriver> | null`). The neighbouring cast on the comparison a few lines up is kept — that one still widens a different generic.
- react-hooks 7 adds the compiler-based `react-hooks/use-memo`, which requires a literal dependency array. `useEcho` and `useEchoNotification` forward a caller-supplied `dependencies` array by design, so the existing disable directive is extended rather than the API changed.

**`@testing-library/vue` dropped as unused**

It was never imported anywhere — the only reference was its own line in `packages/vue/package.json`, and the vue tests use `@vue/test-utils`. It was also the sole reason `@testing-library/dom` was pinned back to 9.x, since 8.1.0 (still the latest published version) hard-pins `^9.3.3`. Removing it takes `chalk@4`, `ansi-styles@4` and `supports-color@7` with it. Vue's 48 tests are unaffected.

**Net effect: 513 → 429 packages.**

Some old transitive versions deliberately remain, because their parents still require them: `ajv@6.15.0` + `json-schema-traverse@0.4.1` (eslint 10 itself still depends on ajv ^6), `minimatch@3.1.5` (`@microsoft/api-extractor`, via `vite-plugin-dts`) and `eslint-visitor-keys@3.4.3` (`@eslint-community/eslint-utils`).

**Verification** — `pnpm build`, `pnpm -r exec vitest --run` (160 passed, 2 skipped — unchanged), `pnpm run lint` (0 errors), `pnpm audit` (clean) and `pnpm run format` (no-op) all exit 0.

`@eslint-react` v5 surfaces 5 new `exhaustive-deps` warnings in `use-echo.ts` (missing `event` and `invokeOnMount` deps). They're warnings, not errors, so CI is green — but they're real findings and fixing them changes re-render behaviour, so they're left for a separate change rather than silently patched here.


taylorotwell:
This one has conflicts. Please mark as ready for review when the requested changes have been made.

pataar:
@taylorotwell Done. Are you okay with me fixing some more dev majors to reduce the vulnerability load? e.g. `vite-plugin-dts`, `vite`, `jsdom` etc. In separate PR's of course.

taylorotwell:
Sure

Issue #453, opened 2025-09-17
Bypass `import.meta.env` type checking on React Native (or possibly other non-Vite projects)


vinzlee97:
### Echo Version

2.2.0

### Laravel Version

12.4.0

### PHP Version

8.2.29

### NPM Version

24.6.0

### Description

Hello team, and thanks for the great package!

This issue might be related to [this](https://github.com/laravel/echo/pull/422#issuecomment-2870522271), but instead of React, it is on React Native project. Since the server uses Reverb, I choose to use Laravel Echo on the client. Based on this [PR](https://github.com/laravel/echo/pull/436), I see this additional `import.meta.env` type check, but on Metro it immediately refuses to acknowledge `import.meta` as it is unsupported.

<img width="661" height="17" alt="Image" src="https://github.com/user-attachments/assets/80274acc-c8f8-4845-b7fc-1934078022fa" />

So my temporary solution is to comment out the defaults which uses `import.meta.env` altogether, then rebuilt the `dist` folder. This works, but I have to do this patch everytime new release comes in.

```diff
    const defaults: ConfigDefaults<BroadcastDriver> = {
        reverb: {
            broadcaster: "reverb",
-            key: import.meta.env.VITE_REVERB_APP_KEY,
-            wsHost: import.meta.env.VITE_REVERB_HOST,
-            wsPort: import.meta.env.VITE_REVERB_PORT,
-            wssPort: import.meta.env.VITE_REVERB_PORT,
-            forceTLS:
-                 (import.meta.env.VITE_REVERB_SCHEME ?? "https") === "https",
+            // key: import.meta.env.VITE_REVERB_APP_KEY,
+            // wsHost: import.meta.env.VITE_REVERB_HOST,
+            // wsPort: import.meta.env.VITE_REVERB_PORT,
+            // wssPort: import.meta.env.VITE_REVERB_PORT,
+            // forceTLS:
+            //     (import.meta.env.VITE_REVERB_SCHEME ?? "https") === "https",
+            forceTLS: true,
            enabledTransports: ["ws", "wss"],
        },
        ...
```

I'm thinking the config can use some kind of flag to indicate whether the defaults can grab from `import.meta.env` (which Vite projects most certainly uses), so that it doesn't have to rely on `import.meta.env`, hence won't do the type checking (e.g. `useImportMetaEnvForDefaults` which defaults to `true` if not provided), but I'm curious about your thoughts on this. Thanks in advance!

### Steps To Reproduce

.

PR #542, opened 2026-08-21
build: upgrade to vite 8 and migrate to unplugin-dts

pataar:
Upgrades the Vite stack: Vite 6 → 8 (Rollup → Rolldown), the dts plugin, Vitest, `@types/node` and jsdom.

## Wins

- **ESM bundles ~33% smaller** from `oxc-minify` — `index.js` 29.1 → 19.6 kB in react/svelte/vue, `echo.js` 22.4 → 14.2 kB.
- **esbuild leaves the dependency tree.** Vite 8 no longer uses it and exposes `transformWithOxc`, which replaces the direct esbuild import in the svelte runes plugin — one transform toolchain instead of two, and one less native binary.
- **A real hole in the published types is fixed.** The shipped `.d.ts` referenced `keyof Events` without ever declaring `Events`, so consumers got `Cannot find name 'Events'` unless they declared it themselves. `unplugin-dts` emits the ambient block the old plugin silently dropped.
- Dead code and deps removed: `@testing-library/svelte` (imported nowhere), `laravel-echo/src/index.iife.ts` (unreferenced — that package's IIFE entry is `src/echo.ts`), and three per-package `overrides` blocks pnpm never read.

## Why `unplugin-dts`, and the one dependency added

`vite-plugin-dts@5` is a thin wrapper whose only dependency is `unplugin-dts@1.0.3`, so going direct removes a layer rather than adding one. Two API changes come with it:

- `rollupTypes` → `bundleTypes`
- **`@microsoft/api-extractor` has to be declared.** It was a transitive *dependency* of `vite-plugin-dts@4` but is an *optional peer* of `unplugin-dts`. This is the one dependency added, and it isn't optional in practice: without it `echo.d.ts` degrades from an 873-line self-contained bundle to 180 lines with relative `./channel` imports — **and the build still exits 0**, so it would ship silently.

## Output verified unchanged

Built `2.x` before and after, then compared:

- Identical file sets and identical runtime export surfaces across all four packages
- `laravel-echo/echo.d.ts` **byte-identical**
- react/svelte/vue `index.d.ts`: **+37 / −0** lines — only the `Events` global above
- Externals unchanged (`pusher-js`, `vue`, `svelte/internal/client`); `laravel-echo` still inlined
- `import.meta.env.VITE_*` rewriting identical

160 tests pass, lint reports 0 errors, `pnpm audit` finds no known vulnerabilities.

## Node support and CI

Node 20 is **not** dropped. `engines` moves from `>=20` to `^20.19.0 || >=22.12.0`, matching Vite 8's floor — the old range admitted 20.0–20.18, which Vite 8 does not support.

The pipelines were the weaker half of this: `tests.yml` and `coding-standards.yml` had **no `setup-node` step at all** and ran on whatever Node `ubuntu-latest` happened to ship. Both now pin it explicitly — `coding-standards` from `.nvmrc`, `tests` across a **22/24 matrix** — with the pnpm cache enabled, matching `publish.yml`.

Node 20 is deliberately absent from that matrix: pnpm 11 requires `>=22.13` and jsdom 30 requires `^22.22.2 || ^24.15.0 || >=26.0.0`, so the dev toolchain cannot run there. `engines` is the consumer contract and devDependency engines don't constrain consumers, so Node 20.19+ consumers stay supported — but worth knowing CI can't exercise that floor.

## Also

- `__dirname` → `import.meta.dirname`, unsupported by the native config loader Vite plans to default to.
- The svelte rune shim started leaking global `$state`/`$effect`/`$derived` into `dist`, where they would collide with Svelte's own declarations. Svelte 5 ships those, so the shim is replaced by a reference to them.
- `autoInstallPeers` is off and the override list is gone, so every peer is declared explicitly.

TypeScript stays on 5.8: 7.0 is blocked both by `@typescript-eslint` (no v9, peer caps at `<6.0.0`) and by `unplugin-dts`, which hard-errors on TS 7 unless `@typescript/typescript6` is installed.


taylorotwell:
Thoughts on this note surfaced agentically?

Blocking — packages/react/vite.config.ts:15, packages/vue/vite.config.ts:15, packages/svelte/vite.config.ts:15: Vite 8 lowers import.meta to {} in the CommonJS bundles before this plugin runs, producing {}.env.VITE_* and causing configureEcho() to throw for CommonJS consumers; transform or preserve these environment accesses in the CJS output.

Please mark as ready for review when the requested changes have been made.

pataar:
> Thoughts on this note surfaced agentically?
> 
> Blocking — packages/react/vite.config.ts:15, packages/vue/vite.config.ts:15, packages/svelte/vite.config.ts:15: Vite 8 lowers import.meta to {} in the CommonJS bundles before this plugin runs, producing {}.env.VITE_* and causing configureEcho() to throw for CommonJS consumers; transform or preserve these environment accesses in the CJS output.
> 
> Please mark as ready for review when the requested changes have been made.

Good one. This was a valid problem. I've mitigated it by marking the variables to be filled by the consumer instead of the local vite. (Validated the output files) This should also partially fix #453.

#453 can be fixed by setting a hard cjs export for React Native. I could fix that in a separate PR if needed.

